PR #35, opened 2014-08-19
https://www.jfrog.com/jira/browse/GAP-158

iNikem:


lqbweb:
That is not completely true. When you have a Maven repository with Unique snapshots, then effectively they will be deployed, even though they are the same.

Sorry, but I agree with iNikem.

jbaruch:
@lqbweb unique snapshots won't be deployed IF they are the same. Maven has a very bad habit of generating different files for the same content (that's because it adds a timestamp into a properties file in META-INF). Once you disable this misfeature everything starts working as expected - similar files won't be uploaded (but you will see a new entry in Artifactory as it was uploaded).

lqbweb:
Yes,  my entries have the same checksum,  but as you say, they appear twice
in artifactory. How can I avoid that?

jbaruch:
I am not sure it should be avoided. Build was performed. New build-info was recorded and generated.Why not to have new entries in Artifactory for that build?

lqbweb:
then, let me the option from gradle to skip the publishing if it is up-to-date (which for me, it sounds like the way to go).

Also, keep in mind, that, it is you who decides when to publish the build-info. What if my build is made out of several build steps that makes impossible to run everything in one gradle go.

PR #43, opened 2014-12-22
https://www.jfrog.com/jira/browse/GAP-162

stormbeta:
Validate all build info tasks have executed before deploying to allow using parallel execution with the artifactory gradle plugin.


lqbweb:
This is actually happening to me as well. Please merge and fix this.

liorhson:
Hey @stormbeta,
Please go to the following [link](http://www.jfrog.com/confluence/display/OSS/How+to+Contribute), and sign the CLA paper, then we would be able to merge you pull request.

lqbweb:
On the latest 4.5 I am still having the same problem, I see some artifactoryPublish tasks getting run after the the plugin deploys everything.

Checking your current source code, it still looks like suffering from the same problem:

`
private boolean isLastTask() {
    return getCurrentTaskIndex() == (getAllArtifactoryTasks().size() - 1);
}
`

That is not right, and it is the same thing as @stormbeta  was trying to solve. The last task may get to run before other tasks are not yet finished. This is what happens with --parallel in gradle.

The original Jira issue was just closed:
https://www.jfrog.com/jira/browse/GAP-162

It seems to me pretty crazy after so many years having the same problem. parallel is not incubating any more....

eyalbe4:
This issue is very old, and it looks like https://github.com/JFrogDev/build-info/pull/134 which has just been merged resolves the issue.
I'm therefore closing this issue.

PR #45, opened 2015-03-09
Add support for incremental build info descriptors via Gradle

noamt:
This modification introduces the 'parent' closure to the Gradle plugin; when configured, all generated build info modules will be appended to the pre-existing descriptor of the build specified in the 'parent' closure.


eyalbe4:
Pull request has been accepted and merged with a few modifications.

Issue #49, opened 2015-05-22
pom file not generated when triggering artifactoryPublish with gradle 2.4


breskeby:
when triggering 

```
gradle clean artifactoryPublish
```

the build fails with gradle 2.4 as the pom file is not yet generated. This is due to changes in the gradle maven-publish plugin. A workaround is to call "gradle generatePomFileForMavenCustomPublication artifactoryPublish" to explicitly call the generatePomFile task for the "MavenCustom" publication. A pullrequest with a fix is on the way

PR #50, opened 2015-05-22
explicitly make artifactoryPublish task depend according pom generati…

breskeby:
this is a fix for issue #49 I just raised to make the gradle artifactory plugin usuable with gradle 2.4


eyalbe4:
Thank you so much Rene!
I'm adding the same for Ivy Publications, as I saw it is also required (see below).
We will soon publish a new release of the plugin
```
            String capitalizedPublicationName = ivyPublication.getName().substring(0, 1).toUpperCase() + ivyPublication.getName().substring(1);
            dependsOn(String.format("generateDescriptorFileFor%sPublication", capitalizedPublicationName));
```

freddy33:
Thanks Eyal.

PR #58, opened 2015-09-22
Handle ant IvyTasks that may be wrapped by Ant presetdef tasks

arothian:
We use ant presetdef tasks to simplify some of our common calls into Ivy's ant tasks. This has the side effect that these ivy calls are wrapped and no longer get detected in Jenkins by the artifactory plugin. This handles the cases in which the IvyTask is wrapped within PreSetDef task.


eyalbe4:
@arothian , thanks a lot for contributing this!
I added another commit on top of your code. Please have a look and confirm the new code works properly on your end.
The following Jira issue tracks this fix:
https://www.jfrog.com/jira/browse/IAP-42

arothian:
@eyalbe4 Your additional changes look good and I tested them in our Jenkins instance against the use case. Thanks for merging!

PR #59, opened 2015-09-22
Deterministic publish order

arothian:
Currently the build extractors for Ivy and Gradle store the resulting list of artifacts to publish in a HashSet. This causes the artifacts to be published in a random order. This updates that to use a TreeSet and provides comparable implementations for related objects so that the published artifacts occur in a deterministic order. This is really helpful for us to determine when a build is complete. This would solve https://www.jfrog.com/jira/browse/GAP-181 in a more predictable manner than is currently implemented.


shikloshi:
Hi @arothian, 
Thanks for this contribution!
I am currently working on testing these changes and will let you know once I'm I will have any conclusions.

Issue #65, opened 2015-10-07
Log verbosely in exception cases within ArtifactoryHttpClient.execute


xasx:
We are currently facing errors like:

```
[ERROR] Failed while reading the response from: PUT https://my.server.address/artifactory/myrepo/org/mynamespace/myartifact/1.0-2015-10-01-r285d831d/myartifact-1.0-2015-10-01-r285d831d.war;vcs.revision=285d831dd064851bb5ef3df45d69b3ee097f3a68;build.timestamp=1443688654812;build.name=Build-My-Artifact;build.number=4711 HTTP/1.1
org.codehaus.jackson.JsonParseException: Unexpected character ('<' (code 60)): expected a valid value (number, String, array, object, 'true', 'false' or 'null')
 at [Source: java.io.StringReader@7759203a; line: 1, column: 2]
	at org.codehaus.jackson.JsonParser._constructError(JsonParser.java:943)
	at org.codehaus.jackson.impl.JsonParserBase._reportError(JsonParserBase.java:632)
	at org.codehaus.jackson.impl.JsonParserBase._reportUnexpectedChar(JsonParserBase.java:565)
	at org.codehaus.jackson.impl.ReaderBasedParser._handleUnexpectedValue(ReaderBasedParser.java:429)
	at org.codehaus.jackson.impl.ReaderBasedParser.nextToken(ReaderBasedParser.java:192)
	at org.codehaus.jackson.map.ObjectMapper._initForReading(ObjectMapper.java:1625)
	at org.codehaus.jackson.map.ObjectMapper._readValue(ObjectMapper.java:1560)
	at org.codehaus.jackson.map.ObjectMapper.readValue(ObjectMapper.java:711)
	at org.codehaus.jackson.JsonParser.readValueAs(JsonParser.java:879)
	at org.jfrog.build.client.ArtifactoryHttpClient.execute(ArtifactoryHttpClient.java:209)
	at org.jfrog.build.extractor.clientConfiguration.client.ArtifactoryBuildInfoClient.tryChecksumDeploy(ArtifactoryBuildInfoClient.java:649)
	at org.jfrog.build.extractor.clientConfiguration.client.ArtifactoryBuildInfoClient.uploadFile(ArtifactoryBuildInfoClient.java:607)
	at org.jfrog.build.extractor.clientConfiguration.client.ArtifactoryBuildInfoClient.deployArtifact(ArtifactoryBuildInfoClient.java:330)
```

[...]

PR #71, opened 2015-11-02
Issue 65 - Add more verbose logs

AlexeiVainshtein:
In case response from the client is not a JSON file, shows the response in the logs.


xasx:
Closes #65

PR #81, opened 2016-04-15
master

MadMatt04:
This modification will cause the teamcity-artifactory-plugin and other tooling based on this project to **not replace** dependency versions in the dependency management section of the pom on artifactory release if those versions are variable-based, such as ${project.version}.

`<dependencyManagement>
<dependencies>
<dependency>
<groupId>com.marand.thinkehr</groupId>
<artifactId>thinkehr-remoting</artifactId>
<version>${project.version}</version>
</dependency>
`
As in, this will be left alone.

It includes a unit test confirming the correct functionality.

Resolves https://www.jfrog.com/jira/browse/TCAP-273

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,3 @@
+language: java
+jdk:
+  - oraclejdk8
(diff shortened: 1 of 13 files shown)


eyalbe4:
This PR is very old. I'm actually not sure why is was left with no updates. It surely needs to rechecked and updated. I'm closing it for now. We can create a new PR for this, if this requirement is still relevant.

PR #100, opened 2016-10-21
Fix pattern matching of path on windows platform

siggae:
https://www.jfrog.com/jira/browse/HAP-823


DimaNevelev:
@siggae  , thank you again for your contribution.
For us to be able to merge your pull request, can you please sign JFrog's CLA?
You can sign through the following link:
http://www.jfrog.com/confluence/display/OSS/How+to+Contribute

rondagostino:
This pull request actually breaks the plugin because it is no longer possible to include a backslash character as part of the regular expression.  For example, when I uploaded with v2.7.2 of the Artifactory plugin (which did not have this change) with this upload spec:

`"pattern": "pom\\.xml"`

I received this output as expected:

`For pattern: pom\.xml 1 artifacts were found`

But in v2.8.0 of the plugin (the first version with this change) the exact same upload spec fails to match and find the file; I get this message instead (note the backslash has been converted to a forward slash)

`For pattern: pom/.xml 0 artifacts were found.`

DimaNevelev:
Thank you for reporting this issue.
We are opened a Jira issue, it be tracked [here](https://www.jfrog.com/jira/browse/BI-380).

PR #112, opened 2017-03-06
conf to record immediate deps only, no transitive

tavin:
We need this in our company b/c we use artifactory to capture "builds" of a particular pom that enumerates versions of various other internal projects as dependencies. We need to be able to grab this list from the artifactory builds API without the transitive dependencies of those projects being included.

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
@@ -429,7 +429,7 @@ public class BuildInfoRecorder extends AbstractExecutionListener implements Buil
             return;
         }
 
-        mergeProjectDependencies(project.getArtifacts());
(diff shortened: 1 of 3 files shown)


tavin:
maybe @eyalbe4 can review this one?

eyalbe4:
@tavin - this one has been waiting for a long time. My apologies for this!
I think that the best approach would be to add an indication in the build-info JSON,  for each dependency, whether is a direct or transitive dependency. We can perhaps add a boolean named **direct** for each dependency.
Please let me know what you think.

tavin:
@eyalbe4 yes, that makes sense! or maybe `transitive: false` instead of `direct: true`. either way it sounds like a good solution.

tavin:
@eyalbe4 is there something I can follow in an issue tracker for this? so I can close my PR.

eyalbe4:
@tavin,
This request hasn't been planned yet. Feel free to open a GitHub issue for this request and close this PR. We'll maek sure to update or link this issue when work on this starts.

tavin:
Resolved by deciding to abandon usage of Artifactory "builds" feature.

Issue #52, opened 2015-07-31
Exception is thown when using Artifactory Plugin in Jenkins


white-speedy:
I am using Artifactory Plugin in Jenkins Maven Project (multi module). While Job is runing NPE is thrown from org.jfrog.build.extractor.maven.BuildInfoRecorder#getExtension. Looking on the code, this is really bug because NULL check is implemented after parameter access .

```
org.apache.maven.InternalErrorException: Internal error: java.lang.NullPointerException
    at org.apache.maven.lifecycle.internal.BuilderCommon.handleBuildError(BuilderCommon.java:142)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:95)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:59)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.singleThreadedBuild(LifecycleStarter.java:183)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute(LifecycleStarter.java:161)
    at org.apache.maven.DefaultMaven.doExecute(DefaultMaven.java:317)
    at org.apache.maven.DefaultMaven.execute(DefaultMaven.java:152)
    at org.jvnet.hudson.maven3.launcher.Maven31Launcher.main(Maven31Launcher.java:132)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:497)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launchStandard(Launcher.java:330)
    at org.codehaus.plexus.classworlds.launcher.Launcher.launch(Launcher.java:238)
    at jenkins.maven3.agent.Maven31Main.launch(Maven31Main.java:181)
    at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
    at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #118, opened 2017-04-13
[HAP-909] NPE on gather build info

nfalco79:
Define the default scope for the artifact dependency in case it's value is null or an empty string.
I've already accept the JFrog CLA (adobe sign)

initial discussion on issue #52 
jira issue [here](https://www.jfrog.com/jira/browse/HAP-909)


eyalbe4:
Thank you @nfalco79!

PR #126, opened 2017-07-14
gradle with -parallel deploying not on last task - GAP-162 

lqbweb:
This was originally resolved by @stormbeta in https://github.com/JFrogDev/build-info/pull/43

But it was never accepted and merged. Strangely, the Jira issue was set to resolved but it was not yet resolved according to the code:

https://www.jfrog.com/jira/browse/GAP-162

Now, I am afraid that the code diverge too much to resolve it under the original pull request, so I followed a slightly different approach following the same style of caching the tasks in an attribute of the class.

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/BuildInfoBaseTask.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/BuildInfoBaseTask.java
@@ -60,6 +60,7 @@ public abstract class BuildInfoBaseTask extends DefaultTask {
     public final Set<GradleDeployDetails> deployDetails = Sets.newTreeSet();
 
     List<BuildInfoBaseTask> artifactoryTasks = null;
+	List<BuildInfoBaseTask> remainingTasks = null;
 
     public abstract void checkDependsOnArtifactsToPublish();
 
@@ -141,7 +142,9 @@ public abstract class BuildInfoBaseTask extends DefaultTask {
      * @return true if this is the last ArtifactoryTask task.
      */
     private boolean isLastTask() {
-        return getCurrentTaskIndex() == (getAllArtifactoryTasks().size() - 1);
+		//        return getCurrentTaskIndex() == (getAllArtifactoryTasks().size() - 1);
+		updateRemainingTasks();
+		return remainingTasks.size() <= 1;
     }
 
     /**
@@ -176,6 +179,24 @@ public abstract class BuildInfoBaseTask extends DefaultTask {
         return artifactoryTasks;
     }
 
+	/**
+     * Analyze the task graph ordered and extract a list of build info tasks
+     *
+     * @return An ordered list of build info tasks
+     */
+    private void updateRemainingTasks() {
+		if(remainingTasks==null) {
+			remainingTasks = new ArrayList<BuildInfoBaseTask>(artifactoryTasks);
+		}
+		List<Task> toRemove = new ArrayList<Task>();
+		for (Task task : remainingTasks) {
+			if(task.getState().getExecuted()) {
+				toRemove.add(task);
+			}
+		}
+		remainingTasks.removeAll(toRemove);
+    }
+	
     public void projectsEvaluated() {
         Project project = getProject();
         if (isSkip()) {


eyalbe4:
@lqbweb,
I reviewed the code in this pull request and it looks like it may fail with parallel execution.
The *remainingTasks* list is not thread safe and a few threads may try to remove items from it at the same time.
Shouldn't making the code inside *getAllArtifactoryTasks()* method synchronized, resolve the issue?

lqbweb:
does it need to be thread safe? there is only one TaskAction in the Task, and the Task will run just once and be different for every project, even with -parallel, it is safe right?

Actually, this makes me wonder why cannot we make "remainingTasks" static and thread safe (same with `artifactoryTasks`). Performance-wise it will be much better, as it will not have to iterate the whole graph all over again for every project. In fact it makes no point at all to cache it in the instance right?

eyalbe4:
@lqbweb,
You're right, these lists are local for a specific task, so thread safety is not an issue.
Do you know what is the reason the existing code fails to find the last task?

lqbweb:
@eyalbe4 yes, your code works on the assumption that the last task will finish after all the previous ones, and that is not neccesarily like that (that is what you do when you check for the current task to be the last index, without having checked for all the previous ones).
So a simple scenario like
:app1, :app2
will not work if they dont depend on each other and :app1 is larger and takes longer to compile than :app2

eyalbe4:
@lqbweb,
Let me know what you think about [this](https://github.com/JFrogDev/build-info/commit/48e671e57f50ba0f679cde5927a81c503088a9d2) commit.
The idea is to let the task with the last index, be in charge of the deployment.
Now however, this task actually waits until all the other tasks confirm that they have finished.
Are you able to build the code and try it out?

lqbweb:
@eyalbe4  it looks ok your solution, I cannot see any flaw, and I have tried. I can build it and give it a try but first confirm me this plz:
is this for sure the approach that you want to follow? Efficiency-wise is better than my solution of course (at the cost of complexity),  but:
- If the last task is blocked until all the others are done, you may also be slowing down the whole build, because there might be more tasks that depend on "artifactoryPublish" that could start moving forward on the last task.
- What is special about the last task in the graph? I think they are just sorted internally by alphabetical order and is very specific from the gradle implementation. Why to delegate to this task the responsibility?
If this is still the approach you want to follow, then I do not see any better solution than yours.
But, if it is ok if the last task to run (and not in the graph order) is the one responsible to deploy, then I think there are much simpler solutions. For instance, you could also use a TaskExecutionGraphListener, or simply update an static counter at the finally{} of the taskAction and deploy when this counter is 1.

eyalbe4:
Thanks for the feedback @lqbweb.
* If there are other tasks that depend on the *artifactoryPublish* task, in most cases at least, it might be actually better to keep them waiting till all of the functionality (artifacts deployment and build-info publication) is completed. 
I guess that there may be scenarios which can slow down the build, for example, if in a specific sub project, there's another task that depends on the "deploying" *artifactoryPublish* task.
* There's no special reason for selecting the task with the last index to be in charge of the deployment, other than then the following: If you're not running the gradle build --parallel, the code will avoid running into the synchronized block in the *isLastTask* method. 
I'm a bit concerned about adding static variables to store data, because they may get us into trouble when running with the gradle daemon (we'll need to reset the statics somehow after the build, while the gradle process is still up and running). 
I agree with you though - there are other ways to implement this solution.

lqbweb:
ok, you are right. After artifactoryPublish it should be as close as possible to published, and in that way it is deterministic which task will perform the deployment.
Let me give it a try, and I will comment you here.

lqbweb:
@eyalbe4 it works fine. Same build time as with my solution. No artifactoryPublish task is run after the deployment. All as expected. Do you know when can we expect a release with this fix?
Thanks!

lqbweb:
@eyalbe4 I am having a dead lock now . I jProfiled it, it is stuck on the wait of isLastTask(), and the rest of the tasks are parked, nothing running. Last notify somehow got lost I think.  Gradle 4.1. Still cannot figure out where the problem is.

eyalbe4:
@lqbweb, thanks for reporting this.
I also can't see the problem in the code...
Let's work together to pinpoint the problem. 
We can add a log print with the task.toString() before and after every "notify", as well as before and after the "wait".
We should then see the task, for which the "notify" gets lost.
Would you like me to add the log prints and share a snapshot version with you? Or you prefer to modify the code and build it?

lqbweb:
Some other projects artifactoryPublish tasks are still in the graph by the time the last isLastTask() hits.
I am printing them out.
These tasks are scheduled to be skipped (printed out as "SKIPPED" in configuration time by gradle), but somehow are still in the graph and when I do task.getState().isSkipped(), they always returns false. At the same time, getState().getExecuted() are false for these tasks (so they have never run).
This does not look like a synchronization issue now.
Any ideas? Maybe we have to check the dependencies up as well?
You can just tell me where do you want to place the prints, or you can do it in a branch and I build it.
Thanks

eyalbe4:
@lqbweb,
If those tasks are never run, will you be able to check if the *taskAction()* method in the *BuildInfoBaseTask* is never executed for these tasks?
If it is not executed, then we may have a solution:
In addition to the existing *finished* property (which we set to to false at the end of the *taskAction()* method), we can try adding a new *started* property, and set it to *true* at the beginning of the method.
Then, in the *getAllArtifactoryTasks()* method, we can just skip tasks which haven't started, and by doing so, no tasks will wait for them.
What do you think?

lqbweb:
yes, that solution I also thought about it. Sounds good. Although, are you 100% sure that the last task in the graph will get to run always after all the others?

Still, I would like to understand what it is happening... Why isSkipped() returns false when they are skipped? maybe isSkipped() is not checking the "onlyIf" of the task? then how is this supposed to be checked in gradle? I think it would be safer if we use the gradle API rather than a new flag (if we find the way of course)

lqbweb:
@eyalbe4  I have checked, as I have set up artifactoryPublish to depend on assemble, when assemble is up to date, artifactory is skipped but the TaskOutcome does not reflect so when querying getState(). Not even in the latest gradle 4.2. 
Also, by reading gradle code, and checking in deep the documentation, I did not find an easy way to infer the skipped status. I think it is kept internally in the TaskRunner.  Therefore, the only alternative to your solution comes by building a reverse dependencies map of the graph and check recursively the dependencies state. I imagine this is a lot of work, so I am afraid the only solution is your flag.

I have posted a question here (in case you want to follow it up):

https://discuss.gradle.org/t/getskipped-does-not-look-up-in-hierarchy/24234

eyalbe4:
Thanks for investigating this further and posting the question @lqbweb!
I'll implement the *started* property solution and let you know when it is pushed and available.

lqbweb:
Have a look to my commit in my repo:

https://github.com/lqbweb/build-info/commit/a8c9fb1e18898c22ffa3e4446ac19556a37e5a75

this does not work. There are further artifactoryPublish tasks called after the deploy :| I am currently researching why.

lqbweb:
@eyalbe4  I found the problem. And you will like it 😄.  I am preparing another pull request. The problem is:

When the last of your tasks in the graph goes to sleep forever is for a good reason, this is, the execution plan contains tasks that should still run. But this execution plan can change while the last task is sleeping. How?
Well, artifactoryPublish depends upon a task (which again did not run by the time the last index task went to sleep), the latter can fail for numerous reasons (in my case throwing an uncaught exception ) and gradle set the automatically the artifactoryPublish to SKIPPED, since its dependency failed.

Therefore, you cannot go to sleep forever. I am looking into the event listeners API of the graph executor to get a notification when this happens.

lqbweb:
@eyalbe4 I would have liked a bit more of support on this topic. But anyway... If you want more of my contribution, better to talk by email (look at my profile).

I have refactored a bit your code :mask:

Sorry, I cannot take your approach, I think it is wrong. Have a look at my commit:

 https://github.com/lqbweb/build-info/commit/618306a5ebda0c8b2c532dea3b667b0156c13e0e

That works fine.

eyalbe4:
@lqbweb,
Thanks for actively continuing to investigate this. I really appreciated this! I'll soon look into your code and we can then continue discussing the best approach.

eyalbe4:
@lqbweb,
I looked at your code. The concept of creating a new task to handle the deployment is indeed interesting.
Does the new code work properly for your projects? In case it does, we can start testing it as well and if all works well, adopt this new approach.

lqbweb:
@eyalbe4 yes, it does work, I have not tested side conditions so it needs a bit more of testing like a dependency task failing (as the finalizedBy block would still run).  I think it needs now to test for the state of the finalizedTask, like indicated here:

https://stackoverflow.com/questions/33285263/gradle-task-that-depends-on-a-failure

eyalbe4:
Sounds good @lqbweb.
When you're ready, feel free to create a pull request.

lqbweb:
@eyalbe4 I have created another pull request, lets continue on the other PR.

eyalbe4:
@lqbweb,
I'm closing this PR, since https://github.com/JFrogDev/build-info/pull/134 is now merged.

PR #134, opened 2017-10-24
Move deploying action to a separated task

lqbweb:
This was raised because of: https://github.com/JFrogDev/build-info/pull/126

The idea of deploying on the last task of the graph gets a bit complex when parallel build is enabled, as we have to deal with waiting for previous tasks to finish, parent dependencies being skipped, triggering errors.

All these complications gradle can already deal with it if you move the deploying action to a task in the rootProject, and you setup dependencies accordingly. 

By doing this I have also detected an incongruity which I needed to fix in order for this to compile:

If I got the idea right, isPublishBuildInfo() (before this diff, in BuildInfoBaseTask.java), should only check the ArtifactoryClientConfiguration, and not the artifactoryPublish task config, as otherwise, the last task specification can decide for the others (and who knows what the last task in the graph is).

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPlugin.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPlugin.groovy
@@ -4,8 +4,10 @@ import org.gradle.api.Project
 import org.jfrog.gradle.plugin.artifactory.dsl.ArtifactoryPluginConvention
 import org.jfrog.gradle.plugin.artifactory.task.ArtifactoryTask
 import org.jfrog.gradle.plugin.artifactory.task.BuildInfoBaseTask
+import org.jfrog.gradle.plugin.artifactory.task.DeployTask
 
 import static org.jfrog.gradle.plugin.artifactory.task.BuildInfoBaseTask.BUILD_INFO_TASK_NAME
+import static org.jfrog.gradle.plugin.artifactory.task.BuildInfoBaseTask.DEPLOY_TASK_NAME
 
 /**
  * @author Lior Hasson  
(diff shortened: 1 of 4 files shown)


eyalbe4:
Thanks a lot for this pull request @lqbweb.
While we are reviewing the code, can you please join our contributors community by signing [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N) so that we can merge your code?

lqbweb:
done. How is the code review going?

eyalbe4:
Great work and thanks a lot for contributing this @lqbweb.
[GAP-280](https://www.jfrog.com/jira/browse/GAP-280) tracks this code change.

rupebac:
@eyalbe4 does it make sense to check for the existence of :artifactoryDeploy? if you dont apply the plugin in the root project it is simply running the artifactoryPublish without running the artifactoryDeploy. Is this the expected behavior?

eyalbe4:
@rupebac,
I've just used [this](https://github.com/eyalbe4/project-examples/tree/master/gradle-examples/gradle-example-multi-repos) project, to see how the plugin behaves, when it is nor applied to the root project.
In this project, the plugin is applied and configured in the api sub-project.
Here's the build output I got with gradle 3.5:
```
$ gradle clean artifactoryP
:api:clean
:shared:clean
:services:webservice:clean UP-TO-DATE
:api:generatePomFileForMavenJavaPublication
:shared:compileJava
:shared:processResources
:shared:classes
:shared:jar
:api:compileJava
:api:processResources NO-SOURCE
:api:classes
:api:jar
:api:artifactoryPublish
:artifactoryPublish
:artifactoryDeploy
Deploying artifact: http://localhost:8081/artifactory/ext-snapshot-local/org/jfrog/test/gradle/publish/api/1.0-SNAPSHOT/api-1.0-SNAPSHOT.jar
Deploying artifact: http://localhost:8081/artifactory/ext-snapshot-local/org/jfrog/test/gradle/publish/api/1.0-SNAPSHOT/api-1.0-SNAPSHOT.properties
Deploying artifact: http://localhost:8081/artifactory/ext-snapshot-local/org/jfrog/test/gradle/publish/api/1.0-SNAPSHOT/api-1.0-SNAPSHOT.pom

BUILD SUCCESSFUL
```
I got a similar output with gradle 4.3.1.
Please let me know what you think.

rupebac:
yes, it is good... My problem was actually that projectsEvaluated was not being called when I was in configure-on-demand (because of #147 ). And therefore it was never applied in the rootProject. I did not think on configure-on-demand by the time I wrote that comment, sorry.

PR #135, opened 2017-10-26
Disable doclint for Java8

ajeans:
The project doesn't build here for Java8, with some javadoc comments failing the build because of doclint.
For example: BuildDependencyPattern triggers a 'bad use of >' for its comment line "libs-release-local:com/goldin/plugins/gradle/0.1.1/*.jar;status+=prod@gradle-plugins :: Build :: Gradle#LATEST => many-jars-build"

Simply disable doclint for now...
See http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -45,6 +45,14 @@ allprojects {
     }
 }
 
+if (JavaVersion.current().isJava8Compatible()) {
+    allprojects {
+        tasks.withType(Javadoc) {
+            options.addStringOption('Xdoclint:none', '-quiet')
+        }
+    }
+}
+
 artifactory {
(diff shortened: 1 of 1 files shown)


eyalbe4:
Thanks for looking into this @ajeans.
We've just pushed [this](https://github.com/JFrogDev/build-info/commit/f67e4362f0d42483087917bb258b102d87e94aa8) commit, which fixes the javadoc errors. This should allow building the code with JDK8.
Please let us know what you think.

ajeans:
Thanks for the quick reply @eyalbe4 
Feel free to close the PR then, it is a much cleaner fix.

PR #147, opened 2018-01-16
Making it work with configureOnDemand

rupebac:
The plugin does not work properly with ConfigureOnDemand.

The callback projectsEvaluated is actually not a good place to do things like "finalizedBy", because in case of using it with configure-on-demand, it gets called when the configuration is already done.

Look at the following SSCCE to reproduce the problem (which is doing the same as you do).
[testGradle.zip](https://github.com/JFrogDev/build-info/files/1635366/testGradle.zip)

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPluginBase.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPluginBase.groovy
@@ -43,7 +43,11 @@ abstract class ArtifactoryPluginBase implements Plugin<Project> {
         addArtifactoryPublishTask(project)
         if (isRootProject(project)) {
             addDeployTask(project)
+        } else {
+            //makes sure the plugin is applied in the root project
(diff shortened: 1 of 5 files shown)


rupebac:
Look at this part of gradle code:

https://github.com/gradle/gradle/blob/d1bbde760a87d85017e7fe3c6da03e52031d96ee/subprojects/core/src/main/java/org/gradle/initialization/DefaultGradleLauncher.java#L252

rupebac:
Just added another commit: As it is now, when configure-on-demand is on, it is forcing the evaluation of all the projects, even if none of their task is initially targeted. By the time :artifactoryDeploy runs, all the other artifactoryPublish must have been run, and therefore already evaluated, if not, it should ignore it.

eyalbe4:
Thanks for this pull request @rupebac!
While we're evaluating and testing your commits, can you please join our contributors community by signing [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N) so that we can merge your code?

rupebac:
ok, just signed it.
btw: I already signed that when my surname was lqbweb :)

eyalbe4:
Thank you for this great contribution @rupebac!
[GAP-166](https://www.jfrog.com/jira/browse/GAP-166) track this change.

PR #152, opened 2018-02-16
Better logging of what went wrong.

mathiasi:
Every now and then I get the "Could not build the build-info object" error in my Jenkins pipeline. However I actually suspect that it is not the building-part that is the problem but rather the publishing. I have no way of telling because the logging will default to "Could not build the build-info object" even if an exception occurs in sendBuildInfo.


eyalbe4:
@mathiasi,
I'd expect to see which method throw the exception in the exception stacktrace.
Any idea why you do not see a stacktrace in the build log?

eyalbe4:
@mathiasi,
I'm closing this one for now, but do let us know if you'd like to continue working on it.

PR #159, opened 2018-03-23
Fix ensureEnds function to use File.separator instead of default "/" …

sendil-d:
…unix separator

- ensureEnds function always look for "/" (unix file separator) even on
  windows. This causes issues in Jenkins when your workspace is T:\

  If you use custom workspace and set to T:\ or some root drive then
  ensureEnds function adds additional "\" resulting in "T:\\".
  getRelativepath which uses this function then ends up adding
  ".." in remote target path. This results in 500 error from artifactory.

  Deploying artifact: https://artifactory.garmin.com/artifactory/Marine-Builds/Regression/ByJob/Mar-Man-openstack-win-TEST/19/../Artifacts/me.txt
  ERROR: remote file operation failed: T:\ at hudson.remoting.Channel@522d50a:os-cw7x-e1c: java.io.IOException: Failed to deploy file. Status code: 500 Response message: Artifactory returned the following errors:
  Path element cannot end with a dot: Marine-Builds/Regression/ByJob/Mar-Man-openstack-win-TEST/19/../ Status code: 500

- This patch addresses this issue by fixing ensureEnds to check for
  passed in separator instead of default "/"

Change-Id: Ib08c01ba235e51b2bb752f929805eee8f4fa4407

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/spec/UploadSpecHelper.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/spec/UploadSpecHelper.java
@@ -452,6 +452,6 @@ public class UploadSpecHelper {
     }
 
     private static String ensureEnds(String s, char endsWith) {
-        return StringUtils.endsWith(s, "/") ? s : (new StringBuilder()).append(s).append(endsWith).toString();
+        return StringUtils.endsWith(s, String.valueOf(endsWith)) ? s : (new StringBuilder()).append(s).append(endsWith).toString();
     }
 }


sendil-d:
How to reproduce the error

1. Jenkins 2.73.2 and configure artifactory plugin in global configuration
2. Latest artifactory plugin
3. Download spec with spec referenced below

```
        {
            "files":
            [
                {
                    "pattern": "Artifacts/**",
                    "target": "M-Builds/Regression/ByJob/$JOB_NAME/$BUILD_NUMBER/",
                    "flat": "False",
                    "explode": "False"
                }
            ]
        }

```
4. Have a shell step with the following script.

```
echo "hello from test job"
if [ -d Artifacts ]; then
  rm -rf Artifacts
fi
mkdir Artifacts
touch Artifacts/me.txt
touch Artifacts/me2.txt
```
5. Run the job to reproduce the error
6. Now rerun the same job with this fix for build-info-extractor library.

eyalbe4:
Thanks for contributing the code @sendil-d!
I added another [commit](https://github.com/JFrogDev/build-info/commit/a1068738354d85b533919739a788850cec3958e7) on top of yours, to further simplify the code.

sendil-d:
cool thanks for improving it further. 

Do you know when this change will be pulled in to https://github.com/JFrogDev/jenkins-artifactory-plugin?

Thanks for your help.

Sendil

eyalbe4:
@sendil-d,
The upcoming Jenkins Artifactory Plugin release will include this fix. We don't yet have a final ETA for the release though.
We can publish a snapshot version which includes this fix if you'd like.

sendil-d:
Ok. Thanks. I will wait for the future release.

Sendil

PR #182, opened 2018-10-10
BI-461 - Race condition in uploadArtifactsBySpec causes a NullPointerException

barbelity:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/ArtifactoryBuildInfoClientBuilder.java
```diff
@@ -0,0 +1,77 @@
...
+
+    private String username;
+    private String password;
+    private String artifactoryUrl;
+    private Log log;
+    private ProxyConfiguration proxyConfiguration;
+    private int connectionTimeout = -1;
+    private int connectionRetry = -1;
+
+    public ArtifactoryBuildInfoClientBuilder() {
```
eyalbe4: There's probably no need to define the constructor.

eyalbe4:
Looks awesome @barbelity!
See my inline comment above.

PR #183, opened 2018-10-10
Add automatic tests in Appveyor

barbelity:


appveyor.yml
```diff
@@ -0,0 +1,21 @@
+stack: jdk 8
+environment:
+  matrix:
+    - APPVEYOR_BUILD_WORKER_IMAGE: Visual Studio 2017
```
yahavi: It is better to use images like here:
https://github.com/jfrog/jfrog-idea-plugin/blob/master/appveyor.yml#L1

appveyor.yml
```diff
@@ -0,0 +1,21 @@
+stack: jdk 8
```
yahavi: Please add also `skip_tags: true`

appveyor.yml
barbelity: We will leave it that way to avoid setting a redundant env_var on the linux machine

appveyor.yml
barbelity: done

PR #202, opened 2019-01-22
BI-471 - PATH environment variable doesn't passed to the npm extractor

yahavi:


build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -16,10 +18,16 @@
 public class CommandExecutor implements Serializable {
     private static final long serialVersionUID = 1L;
 
+    private String[] env;
     private String executablePath; // Path to an executable file
```
eyalbe4: Let's change to above comment to the javadoc format.

eyalbe4:
👍 @yahavi 
See my inline suggestion.

PR #209, opened 2019-02-13
Replace Guava with java 8 utils

RobiNino:


eyalbe4:
LIKE @RobiNino!
Just one comment - let's add unit tests for the CommonUtils methods.
Thanks!

PR #213, opened 2019-02-19
Set/Delete properties - fix

RobiNino:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/ArtifactorySearcher.java
```diff
@@ -27,26 +27,30 @@
         this.log = log;
     }
 
-    List<AqlSearchResult.SearchEntry> SearchByFileSpec(FileSpec file) throws IOException{
+    List<AqlSearchResult.SearchEntry> SearchByFileSpec(FileSpec file) throws IOException {
         AqlSearchHelper aqlHelper = new AqlSearchHelper(client);
         WildcardsSearchHelper wildcardHelper = new WildcardsSearchHelper(client);
-
+        List<AqlSearchResult.SearchEntry> results = new ArrayList<>();
```
eyalbe4: It looks like we're creating here a new ArrayList, which will most likely not be used. The creation of the list has a cost.

PR #223, opened 2019-03-24
Add required changes to support the new eclipse plug-in

yahavi:


build-info-extractor/src/main/java/org/jfrog/build/extractor/scan/Issue.java
```diff
@@ -0,0 +1,121 @@
...
+        }
+        Issue otherIssue = (Issue) other;
+        if (StringUtils.isEmpty(component)) {
+            return StringUtils.equals(description, otherIssue.description) && StringUtils.equals(summary, otherIssue.summary);
+        }
+        return StringUtils.equals(component, otherIssue.component) && StringUtils.equals(summary, otherIssue.summary);
+    }
+
+    @Override
+    public int hashCode() {
```
eyalbe4: We can replace the hashcode implementation here with *Objects.hash(...).*.

build-info-extractor/src/main/java/org/jfrog/build/extractor/scan/Artifact.java
```diff
@@ -0,0 +1,60 @@
...
+    private GeneralInfo generalInfo = new GeneralInfo();
+    private Set<Issue> issues = new HashSet<>();
+    private Set<License> licenses = new HashSet<>();
+
+    // Empty constructor for serialization
+    public Artifact() {
+    }
+
+    public Artifact(GeneralInfo generalInfo, Set<Issue> issues, Set<License> licenses) {
+        this.generalInfo = new GeneralInfo(generalInfo);
```
eyalbe4: Question - wouldn't a simple assignment be sufficient here?  - 
```this.generalInfo = generalInfo;```
Do we really need a clone of GeneralInfo here?
If so, then the constructor in GeneralInfo which accepts another GeneralInfo may be redundant.

The same question goes for *this.issues* and *this.licenses* in the same method - are we actually aggregating issues and licenses (with addAll) or this is a one time assignment?

File: build-info-extractor/src/main/java/org/jfrog/build/extractor/scan/DependenciesTree.java [hunk 3]
```diff
@@ -0,0 +1,117 @@
...
+import javax.swing.tree.DefaultMutableTreeNode;
+import java.util.Comparator;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.Vector;
+
+/**
+ * @author yahavi
+ */
+public class DependenciesTree extends DefaultMutableTreeNode {
```
eyalbe4: I suggest adding a description for class, explaining what it is for. This could help the random code reader underatand the context.

PR #224, opened 2019-04-08
Artifactory Jenkins log doesn't display NPM log output

yahavi:


build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
```diff
@@ -37,9 +37,9 @@ public boolean isNpmInstalled() {
         }
     }
 
-    public void install(File workingDirectory, List<String> extraArgs) throws IOException {
+    public String install(File workingDirectory, List<String> extraArgs) throws IOException {
         try {
-            runCommand(workingDirectory, new String[]{"i"}, extraArgs);
+            return runCommand(workingDirectory, new String[]{"i"}, extraArgs);
```
barbelity: We are missing error output here, it may be better to output also the stderr messages.

PR #232, opened 2019-06-02
BI-486 - NpmDriver fails when executed with npm's --verbose flag

eyalbe4:


build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
```diff
@@ -72,12 +73,26 @@ public String version(File workingDirectory) throws IOException, InterruptedExce
     }
 
     public String configList(File workingDirectory, List<String> extraArgs) throws IOException, InterruptedException {
-        return runCommand(workingDirectory, new String[]{"c", "ls", "--json"}, extraArgs);
+        // We're making sure that the --verbose npm option is not included in the options list,
+        // because it will break the the command output parsing.
+        // Remove the --verbose option.
+        List<String> args = new ArrayList<>();
+        for (String arg : extraArgs) {
+            if (!"--verbose".equals(arg)) {
```
yahavi: Please take into consideration other verbose variations:
Info: `-d`, `--loglevel info`
Verbose: `-dd`, `--loglevel verbose`
Silly: `-ddd`, `--loglevel silly`

yahavi: Maybe appending `-s` to `npm config ls` command would be a better approach

PR #234, opened 2019-06-18
Remove wrapper task

rahulsom:
With newer versions of gradle, that task is already available. This causes errors in building.

Also, the wrapper currently uses 4.8 whereas this task tried to use 2.4

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -536,7 +536,3 @@ idea {
         }
     }
 }
-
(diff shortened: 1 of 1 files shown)


eyalbe4:
Thank you @rahulsom!
[BI-493](https://www.jfrog.com/jira/browse/BI-493) tracks this fix.

PR #243, opened 2019-08-29
add issues collection support

RobiNino:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryBuildInfoClient.java
```diff
@@ -81,6 +79,8 @@
     public static final String APPLICATION_VND_ORG_JFROG_ARTIFACTORY_JSON = "application/vnd.org.jfrog.artifactory+json";
     public static final String APPLICATION_JSON = "application/json";
     public static final String ITEM_LAST_MODIFIED = "/api/storage/";
+    private static final String LATEST = "LATEST";
```
yahavi: Looks like LATEST and LAST_RELEASE are not in use

build-info-api/src/main/java/org/jfrog/build/api/IssueTracker.java
```diff
@@ -37,4 +38,22 @@ public String getVersion() {
...
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof IssueTracker)) {
+            return false;
+        }
+
+        IssueTracker that = (IssueTracker) o;
+
+        if (!Objects.equals(name, that.name)) {
```
yahavi: Can be simplified to: 
return (Objects.equals(name, that.name) && Objects.equals(version, that.version));

build-info-api/src/main/java/org/jfrog/build/api/Vcs.java
```diff
@@ -41,6 +42,31 @@ public String toString() {
...
+        if (this == o) {
+            return true;
+        }
+        if (!(o instanceof Vcs)) {
+            return false;
+        }
+
+        Vcs that = (Vcs) o;
+
+        if (!Objects.equals(revision, that.revision)) {
```
yahavi: Can be simplified to:
return (Objects.equals(revision, that.revision) && Objects.equals(url, that.url));

build-info-api/src/main/java/org/jfrog/build/api/IssuesCollectionConfig.java
```diff
@@ -0,0 +1,118 @@
+package org.jfrog.build.api;
+
+import org.apache.commons.lang.StringUtils;
+
+import java.io.IOException;
+import java.io.Serializable;
+
+public class IssuesCollectionConfig implements Serializable {
```
eyalbe4: I suggest adding a description for class, describing its role.

build-info-api/src/main/java/org/jfrog/build/api/IssuesCollectionConfig.java
```diff
@@ -0,0 +1,118 @@
...
+        }
+        if (issues.keyGroupIndex == 0) {
+            throw new IOException(ISSUES_COLLECTION_ERROR_PREFIX + MISSING_CONFIGURATION_ERROR + "keyGroupIndex");
+        }
+        if (issues.summaryGroupIndex == 0) {
+            throw new IOException(ISSUES_COLLECTION_ERROR_PREFIX + MISSING_CONFIGURATION_ERROR + "summaryGroupIndex");
+        }
+    }
+
+    public static class Issues implements Serializable {
```
eyalbe4: I suggest adding a description for each of this class fields.

build-info-api/src/main/java/org/jfrog/build/api/Vcs.java
```diff
@@ -41,6 +42,26 @@ public String toString() {
                 ", url='" + url + '\'';
     }
 
+    @Override
+    public boolean equals(Object o) {
```
eyalbe4: Is this method still needed, following the changes in the Jenkins Plugin code base?

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryBuildInfoClient.java
```diff
@@ -190,6 +188,11 @@ public void sendBuildInfo(String buildInfoJson) throws IOException {
     }
 
     public Build getBuildInfo(String buildName, String buildNumber) throws IOException {
+        buildNumber = getLatestBuildNumberFromArtifactory(buildName, buildNumber);
```
eyalbe4: I suggest adding the following comment above this line:
// Only If the value of the buildNumber sent is "LATEST" or "LAST_RELEASE", replace the value with a specific build number.

build-info-extractor/src/main/java/org/jfrog/build/extractor/issuesCollection/IssuesCollector.java
```diff
@@ -0,0 +1,147 @@
...
+        Set<Issue> affectedIssues = new HashSet<>();
+
+        Pattern pattern = Pattern.compile(issuesConfig.getIssues().getRegexp());
+        String[] lines = gitLog.split("\\R");
+        for (String line : lines) {
+            Matcher matcher = pattern.matcher(line);
+            if (matcher.find()) {
+                Issue foundIssue = getMatchingIssue(keyIndex, summaryIndex, matcher, issuesConfig);
+                affectedIssues.add(foundIssue);
+                logger.debug("Found issue: " + foundIssue.getKey());
```
eyalbe4: I suggest changing this from *debug* to *info* and change the message to:
"Added issue XXX to the build-info."

PR #244, opened 2019-09-02
Add git utils to extract Vcs details from dot git

RobiNino:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/GitUtils.java
```diff
@@ -0,0 +1,143 @@
...
+            return dotGit;
+        }
+        return getDotGit(file.getParentFile());
+    }
+
+    /**
+     * Extract Vcs details from .git configuration.
+     */
+    public static Vcs extractVcs(File workingDir, Log log) throws IOException {
+        File dotGit = getDotGit(workingDir);
```
eyalbe4: I think that a NullPointerException can be thrown here, if there's no .git directory (the *getDotGit* method can return null). I suggest throwing a FileNotFoundException with a clear message here.

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/GitUtils.java
```diff
@@ -0,0 +1,143 @@
...
+            originalUrl += ".git";
+        }
+
+        String maskedUrl = maskCredentialsInUrl(originalUrl);
+        log.debug("Fetched url from git config: " + maskedUrl);
+        return maskedUrl;
+    }
+
+    static String maskCredentialsInUrl(String originalUrl) throws IOException {
+        Pattern pattern = Pattern.compile(CREDENTIALS_IN_URL_REGEXP);
```
eyalbe4: Compiling a reg exp is a pretty expensive operation. I suggest we do it only once, and save the compiled pattern.

PR #245, opened 2019-09-02
Issue collection improvements

RobiNino:


build-info-api/src/main/java/org/jfrog/build/api/IssuesCollectionConfig.java
```diff
@@ -47,12 +51,19 @@ public void validateConfig() throws IOException {
...
         private String regexp;
+        // The capturing group index in the regular expression used for retrieving the issue key.
         private int keyGroupIndex;
+        // The capturing group index in the regular expression for retrieving the issue summary.
         private int summaryGroupIndex;
+        // The issue tracking URL. This value is used for constructing a direct link to the issues in the Artifactory build UI.
         private String trackerUrl;
+        // Set to true, if you wish all builds to include issues from previous builds.
         private boolean aggregate;
+        // If aggregate is set to true, this property indicates how far in time should the issues be aggregated.
```
eyalbe4: Since the above descriptions are for class members, I suggest using the standard javadoc description annotation:
```
/**
 * If aggregate is set to true, this property indicates how far in time should the issues be aggregated.
 */
```
instead of
```
// If aggregate is set to true, this property indicates how far in time should the issues be aggregated.
``` 

Same goes for the *version* member description at the top of the file.

PR #250, opened 2019-09-06
Replace deprecated method use

ljacomet:
Pretty straightforward as it effectively [inline the code](https://github.com/gradle/gradle/blob/28d8f6d3372993d2b2212161c85c6ddf797ceb20/subprojects/maven/src/main/java/org/gradle/api/publish/maven/internal/publisher/MavenNormalizedPublication.java#L73-L76) from the deprecated method in Gradle.


eyalbe4:
Thanks for this @ljacomet!
@yahavi is currently working on gradle 6 compatibility of the plugin. He'll soon test your suggested fix as part of this task.

Issue #255, opened 2019-09-12
GradleBuildInfoExtractor triggers deprecation warnings when using Gradle 5.6 and building in parallel


wilkinsona:
When building in parallel, `GradleBuildInfoExtractor` causes numerous deprecation warnings to be logged when it is called as part of the `artifactoryDeploy` task. The warnings are all similar to the following:

> The configuration :spring-aop:annotationProcessor was resolved without accessing the project in a safe manner. This may happen when a configuration is resolved from a thread not managed by Gradle or from a different project. See https://docs.gradle.org/5.6.2/userguide/troubleshooting_dependency_resolution.html#sub:configuration_resolution_constraints for more details. 
This behaviour has been deprecated and is scheduled to be removed in Gradle 6.0.

The stack trace of the code that triggers the warning is always similar to the following:

```
at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.resolveToStateOrLater(DefaultConfiguration.java:566)
at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration.getResolvedConfiguration(DefaultConfiguration.java:556)
at org.gradle.api.internal.artifacts.configurations.DefaultConfiguration_Decorated.getResolvedConfiguration(Unknown Source)
at org.jfrog.gradle.plugin.artifactory.extractor.GradleBuildInfoExtractor.calculateDependencies(GradleBuildInfoExtractor.java:330)
at org.jfrog.gradle.plugin.artifactory.extractor.GradleBuildInfoExtractor.extractModule(GradleBuildInfoExtractor.java:296)
at org.jfrog.gradle.plugin.artifactory.extractor.GradleBuildInfoExtractor.extract(GradleBuildInfoExtractor.java:119)
at org.jfrog.gradle.plugin.artifactory.task.DeployTask.prepareAndDeploy(DeployTask.java:148)
at org.jfrog.gradle.plugin.artifactory.task.DeployTask.collectProjectBuildInfo(DeployTask.java:50)
at org.jfrog.gradle.plugin.artifactory.task.DeployTask.taskAction(DeployTask.java:44)
at java.lang.reflect.Method.invoke(Method.java:498)
at org.gradle.internal.reflect.JavaMethod.invoke(JavaMethod.java:103)
at org.gradle.api.internal.project.taskfactory.StandardTaskAction.doExecute(StandardTaskAction.java:49)
[...]

PR #257, opened 2019-09-16
GradleBuildInfoExtractor triggers deprecation warnings when using Gradle 5 with --parallel

yahavi:
#255


eyalbe4:
@yahavi - I reverted the code of this PR with this commit - https://github.com/jfrog/build-info/commit/68a5dc9d7f5b823077a51025dba198579176eee0 because this code seems to be creating an issue, following https://github.com/jfrog/build-info/pull/261

PR #260, opened 2019-09-22
BI-498 - GitUtils : Look for revision in packed-ref in case ref is no…

liron-shalom:
…t exists.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/GitUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/GitUtils.java
@@ -97,13 +97,28 @@ public class GitUtils {
 
         // Else, if found ref try getting revision using it
         File pathToRef = new File(dotGit, revisionOrRef.ref);
-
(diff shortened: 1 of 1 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/GitUtils.java
```diff
@@ -97,13 +97,28 @@ private static String extractVcsRevision(File dotGit, Log log) throws IOExceptio
...
         // Else, if found ref try getting revision using it
         File pathToRef = new File(dotGit, revisionOrRef.ref);
-
-        // Read HEAD file for ref or revision
-        try (BufferedReader br = new BufferedReader(new FileReader(pathToRef))) {
-            String line;
-            while ((line = br.readLine()) != null) {
-                log.debug("Fetched revision from git config: " + line.trim());
-                return line.trim();
+        if (pathToRef.exists()) {
```
eyalbe4: I suggest we add a unit test specifically for the *extractVcsRevision*, to test the two possible scenarios.

PR #268, opened 2019-10-24
BI-502 - Artifactory Dependencies Client - encode props with URLEncoder

liron-shalom:
This encoder allows using special characters.


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryDependenciesClient.java
```diff
@@ -217,12 +222,26 @@ private HttpResponse executeDownload(String artifactUrl, boolean isHead, Map<Str
...
-        String url = ArtifactoryHttpClient.encodeUrl(urlPath + "?properties=" + props);
+        String url = ArtifactoryHttpClient.encodeUrl(urlPath + "?properties=");
+        // Use URLEncoder to allow special characters use
+        url += encodeProps(props);
         PreemptiveHttpClient client = httpClient.getHttpClient();
         HttpPut httpPut = new HttpPut(url);
         checkNoContent(client.execute(httpPut), "Failed to set properties to '" + urlPath + "'");
     }
 
+    private String encodeProps(String props) throws UnsupportedEncodingException {
```
eyalbe4: The code already includes a method which encodes properties (as matrix params).
The method is used by the maven and gradle upload flows, but there's a chance it is also used by generic upload with file spec  See the *encodePath* method in the *DeploymentUrlUtils* class.
One of the methods should probably replace the other, or the code should be merged, depending on the advantages of each API.

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryDependenciesClient.java
```diff
@@ -217,12 +221,23 @@ private HttpResponse executeDownload(String artifactUrl, boolean isHead, Map<Str
...
     public void setProperties(String urlPath, String props) throws IOException {
-        String url = ArtifactoryHttpClient.encodeUrl(urlPath + "?properties=" + props);
+        String url = ArtifactoryHttpClient.encodeUrl(urlPath + "?properties=");
+        url += DeploymentUrlUtils.buildMatrixParamsString(mapPropsString(props));
         PreemptiveHttpClient client = httpClient.getHttpClient();
         HttpPut httpPut = new HttpPut(url);
         checkNoContent(client.execute(httpPut), "Failed to set properties to '" + urlPath + "'");
     }
 
+    private ArrayListMultimap<String, String> mapPropsString(String props) throws UnsupportedEncodingException {
```
eyalbe4: I suggest adding some integration tests, to check set props in general, including props with special characters.
I suggest that setting props with generic upload is also tested.
See the *IntegrationTestsBase* class, which is the base class of the build-info integration tests.

build-info-extractor/src/test/java/org/jfrog/build/extractor/clientConfiguration/util/spec/SpecsHelperIntegrationTest.java
```diff
@@ -44,6 +46,47 @@ protected void cleanup() throws IOException {
         deleteContentFromRepo(localRepo);
     }
 
+    @Test
+    public void propsTest() throws Exception {
```
eyalbe4: I like this test method, but I think that it should not be part of this class.
This class uses the "test cases" infrastructure, which executes multiple tests (see the *integrationTests* method in the same class).
So there are probably two options. Move this method into a new class (see the *DownloadTest* class for example) or replace this test with additional test cases.

PR #261, opened 2019-09-26
Allow parallel upload of artifacts

facewindu:
While helping customers and OSS projects to optimize the performance of their build, we (Gradle, Inc.) noticed that the `artifactoryDeploy` task was very long to execute.
This is because uploads are executing serially.

This PR allows to upload artifacts in parallel.
The number of forked threads defaults to 8, but can be customized using the `publishForkCount` in the `artifactory{}` DSL extension.

See for instance the net results in the https://github.com/spring-projects/spring-framework OSS project:

- before this PR:
  - https://ge.spring.io/s/ju7kyr3p4m62c/timeline?task=jb2ueukhvhszg
  - `artifactoryDeploy` needs 1 minute to perform 89 uploads

- with this PR:
  - https://ge.spring.io/s/e65kj475jjspk/timeline?task=jb2ueukhvhszg
  - `artifactoryDeploy` needs 7 seconds to perform 89 uploads

PR #270, opened 2019-10-24
Gradle - Remove 'The configuration was resolved without accessing the project in a safe manner' warning

yahavi:
Handle this kind of warnings by using Gradle's Mutex mechanism:

> The configuration :services:webservice:annotationProcessor was resolved without accessing the project in a safe manner.  This may happen when a configuration is resolved from a thread not managed by Gradle or from a different project.  See https://docs.gradle.org/5.6.3/userguide/troubleshooting_dependency_resolution.html#sub:configuration_resolution_constraints for more details. This behaviour has been deprecated and is scheduled to be removed in Gradle 6.0.

Resolves also: https://github.com/jfrog/build-info/issues/255

Guard `configuration.getResolvedConfiguration()` and remove parallelism when populating the build info object.
Uploads still occur in parallel, as contributed in https://github.com/jfrog/build-info/pull/261.

@facewindu
Thanks a lot for making the uploads parallel.
`getMutationState().withMutableState()` freezes when running in non-Gradle threads. To avoid it, we removed parallelism when populating the build info object. As mentioned previously, uploads to Artifactory will still be in parallel.
Can you please take a look? It will be highly appreciated!

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
@@ -372,7 +372,7 @@ public class GradleBuildInfoExtractor implements BuildInfoExtractor<Project> {
                 log.info("Artifacts for configuration '{}' were not all resolved, skipping", configuration.getName());
                 continue;
             }
-            ResolvedConfiguration resolvedConfiguration = configuration.getResolvedConfiguration();
+            ResolvedConfiguration resolvedConfiguration = getResolvedConfiguration(project, configuration);
             Set<ResolvedArtifact> resolvedArtifactSet = resolvedConfiguration.getResolvedArtifacts();
             for (final ResolvedArtifact artifact : resolvedArtifactSet) {
                 File file = artifact.getFile();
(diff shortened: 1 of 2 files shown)


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
```diff
@@ -143,17 +113,12 @@ public Build extract(Project rootProject) {
...
-            try {
-                ExecutorService executor = Executors.newFixedThreadPool(publishForkCount);
-                CompletableFuture<Void> allModules = CompletableFuture.allOf(allProjects.stream()
-                        .map(project -> CompletableFuture.runAsync(() -> addModule(bib, project), executor))
-                        .toArray(CompletableFuture[]::new));
-                allModules.get();
-            } catch (InterruptedException | ExecutionException e) {
-                throw new RuntimeException(e);
+        for (Project project : allProjects) {
+            if(project.getState().getExecuted()) {
```
eyalbe4: if(project. --> if (project.

ljacomet:
@yahavi @eyalbe4 Hey folks, would you consider joining us on the Gradle community slack so we can discuss how we could help getting the best of both worlds: performance and Gradle correctness? Invite link: https://gradl.es/slack-invite-2019

PR #265, opened 2019-10-09
Publication of all artifacts for Ivy and Maven

ljacomet:
This pull request addresses the issue raised in #247 for both Ivy and Maven.

PR #271, opened 2019-10-28
Add publishing of metadata (.module file) to the plugin - Gradle 4 fix

yahavi:
Fix the following exception in Gradle 4:
> Caused by: java.lang.NoSuchMethodError: org.gradle.api.publish.maven.internal.publisher.MavenNormalizedPublication.getAdditionalArtifacts()Ljava/util/Set;
        at org.jfrog.gradle.plugin.artifactory.task.helper.TaskHelperPublications.getArtifactDeployDetails(TaskHelperPublications.java:257)

@ljacomet the exception introduced by the following change:  https://github.com/jfrog/build-info/pull/265/commits/04e8afd9ae4550ef2b04d8f89a3e91c0f1bd49b5#diff-dd7b7a4cca93e9ac769933a1811d39dcR255. 
Can you please take a look on this pull request? Thanks!

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/helper/TaskHelperPublications.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/helper/TaskHelperPublications.java
@@ -248,13 +248,21 @@ public class TaskHelperPublications extends TaskHelper {
                     addMavenArtifactToDeployDetails(deployDetails, publicationName, builder, artifactInfo, mavenPublication);
                 }
             }
-            // Second adding the main artifact of the publication, if present
-            if (mavenNormalizedPublication.getMainArtifact() != null) {
-                createPublishArtifactInfoAndAddToDeployDetails(mavenNormalizedPublication.getMainArtifact(), deployDetails, mavenPublication, publicationName);
+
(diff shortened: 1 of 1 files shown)


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/helper/TaskHelperPublications.java
```diff
@@ -248,13 +248,21 @@ public boolean hasModules() {
...
+            try {
+                // Gradle 5.0 and above:
+                artifacts = mavenNormalizedPublication.getAdditionalArtifacts();
+                // Second adding the main artifact of the publication, if present
+                if (mavenNormalizedPublication.getMainArtifact() != null) {
+                    createPublishArtifactInfoAndAddToDeployDetails(mavenNormalizedPublication.getMainArtifact(), deployDetails, mavenPublication, publicationName);
+                }
+            } catch (NoSuchMethodError error) {
+                // Compatibility with older versions of Gradle:
+                artifacts = mavenNormalizedPublication.getAllArtifacts();
```
ljacomet: This collection will again contain the Maven POM artifact.
So it needs to be filtered again in the code below, especially as this plugin has an option to not publish the POM file.

PR #275, opened 2019-11-05
Reintroduce buildinfo computation parallelism

facewindu:
Discussing it internally within Gradle, this PR reintroduces parallel computation of the build infos.

  Doing so in non-Gradle-managed threads is in theory unsafe, and Gradle logs a warning about that.
  To work around this:
    - use internal Gradle APIs to query only configuration in state ARTIFACTS_RESOLVED
    - Silence Gradle's warning, as what we do is actually read-only, and safe without locking the project state

This effectively replaces https://github.com/jfrog/build-info/pull/270

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
@@ -27,12 +27,32 @@ import org.gradle.api.artifacts.Configuration;
 import org.gradle.api.artifacts.ModuleVersionIdentifier;
 import org.gradle.api.artifacts.ResolvedArtifact;
 import org.gradle.api.artifacts.ResolvedConfiguration;
-import org.gradle.api.internal.project.ProjectInternal;
+import org.gradle.api.internal.artifacts.configurations.ConfigurationInternal;
+import org.gradle.api.internal.artifacts.configurations.DefaultConfiguration;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.tasks.TaskState;
-import org.jfrog.build.api.*;
-import org.jfrog.build.api.builder.*;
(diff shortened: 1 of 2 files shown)


facewindu:
@ljacomet can you sanity check ?
@yahavi please review and discuss if needed

facewindu:
Hi  @yahavi 
Thank you for the reproducer. I'll look as soon as I can next week, and keep you posted.

ghale:
@yahavi I've been looking into this issue and a couple of things are blocking me.  You mention that it should work with Gradle 4, but the example project without any changes uses Gradle 4 and pulls the latest 4.x version from jcenter (`4.+`).  If I try to run that, I get a `java.lang.NoClassDefFoundError` on and internal class (`org/gradle/api/publish/internal/PublicationArtifactSet`).  So it seems like the plugin is already not Gradle 4 compatible and likely couldn't be unless the dependency on the internal class is removed.

Secondly, if I try to run that example project with Gradle 5.6.4 and @facewindu's changes, I can't reproduce the issue reported above (`IllegalStateException`).  This is confusing as it seems like this would be a determinate error (i.e. it shouldn't sometimes happen and sometimes not).  Can you provide some further guidance on reproducing the issue?

yahavi:
@ghale Thanks for looking into this.
1. I just ran `clean artifactoryPublish` with Gradle 4.10.3 and macOS and it worked. Which version do you have?
2. Please run `clean artifactoryPublish` with Gradle 6.

And if something is not clear yet, please don't hesitate to ask more questions.

ghale:
Ok, I was just using the wrapper in the example project, which uses 4.3.1.  I get the same error you reported now. Am I correct in assuming that Gradle 4 compatibility is limited to the latest 4.x version and not all 4.x versions? 

And I also get the `IllegalStateException` with Gradle 6.  For some reason, I thought you were reporting it with Gradle 5 - my mistake.

Thanks for the clarification.

ghale:
I've submitted a separate PR (https://github.com/jfrog/build-info/pull/287) which addresses the issues reported above and supersedes this PR.

facewindu:
closing this as superseded

PR #277, opened 2019-11-11
Add access token support in Artifactory's various clients

liron-shalom:


build-info-client/src/main/java/org/jfrog/build/client/ArtifactoryHttpClient.java
```diff
@@ -60,19 +61,25 @@
...
     private final String password;
+    private final String accessToken;
     private int connectionTimeout = DEFAULT_CONNECTION_TIMEOUT_SECS;
     private int connectionRetries = DEFAULT_CONNECTION_RETRY;
     private ProxyConfiguration proxyConfiguration;
 
     private PreemptiveHttpClient deployClient;
 
-    public ArtifactoryHttpClient(String artifactoryUrl, String username, String password, Log log) {
+    public ArtifactoryHttpClient(String artifactoryUrl, String username, String password, String accessToken, Log log) {
```
eyalbe4: Any reason why we need a constructor that gets both username/password and an access token? This is reflected also in other constructors in the code.

build-info-client/src/main/java/org/jfrog/build/client/ArtifactoryHttpClient.java
```diff
@@ -278,4 +285,10 @@ private String getResponseEntityContent(HttpEntity responseEntity) throws IOExce
         }
         return "";
     }
+
+    public void addAccessTokenHeaderToRequestIfNeeded(AbstractHttpMessage httpRequest) {
```
eyalbe4: This method seems to be invoked from many places in the code. This is different though when setting the username and password. In other words, when using a username/password, we don't need to worry about setting them from each and every request. This approach is easier to maintain and reduces the risk for a bug.
Have a look at the *PreemptiveHttpClient* class, which includes the core http client functionality.

PR #285, opened 2019-12-01
BI-510 - Add support for usage-report to Artifactory

barbelity:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryBaseClient.java
```diff
@@ -149,4 +157,44 @@ public boolean isRepoExist(String repo) throws IOException {
...
+        HttpResponse httpResponse = null;
+        try {
+            httpResponse = httpClient.getHttpClient().execute(request);
+            StatusLine statusLine = httpResponse.getStatusLine();
+            if (statusLine.getStatusCode() != HttpStatus.SC_OK) {
+                throw new IOException(String.format("Artifactory response: %s %s", statusLine.getStatusCode(), statusLine.getReasonPhrase()));
+            }
+        } finally {
+            // We are using the same client for multiple operations therefore we need to restore the connectionRetries configuration.
+            httpClient.setConnectionRetries(connectionRetries);
```
eyalbe4: Looks like there's no reason to "restore" the value of *connectionRetries* value, because it never changes after it is initialized in the PreemptiveHttpClient constructor.
I suggest removing the *setConnectionRetries* method, to make sure no one uses it.

build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
```diff
@@ -0,0 +1,51 @@
+package org.jfrog.build.extractor.usageReport;
+
+/**
+ * Created by Bar Belity on 20/11/2019.
+ */
+public class UsageReporter {
```
eyalbe4: I suggest renaming this class from *UsageReporter* to *UsageReport*, because it is merely a POJO. It doesn't actually perform the reporting (so it is not the reporter).

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryBaseClient.java
barbelity: Since this method is used by the ArtifactoryBaseClient, I didn't remove it completely.
I did remove all unnecessary usages of it.

PR #291, opened 2019-12-23
Add Support for Go build and publish.

liron-shalom:


build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoCommand.java
```diff
@@ -0,0 +1,109 @@
...
+abstract class GoCommand implements Serializable {
+
+    protected static final String SHA1 = "SHA1";
+    protected static final String MD5 = "MD5";
+    protected static final String LOCAL_GO_MOD_FILENAME = "go.mod";
+    private static final long serialVersionUID = 1L;
+    private static final ArtifactoryVersion MIN_SUPPORTED_ARTIFACTORY_VERSION = new ArtifactoryVersion("6.10.0");
+
+    ArtifactoryClientBuilderBase clientBuilder;
+    ArtifactoryBaseClient client;
```
eyalbe4: We shouldn't probably manage in this class both ArtifactoryClientBuilderBase and ArtifactoryBaseClient instances. This makes the decision of which one to use harder. We should simplify the implementation by maintaining only one.
In addition, do we really need to store the base class ArtifactoryBaseClient and not ArtifactoryBuildInfoClient? This can save us a few castings in the code (castings should generally be avoided when possible).method 

Also notice that the **preparePrerequisites(deploymentRepo, client)** receives the client as an argument, although it is accessible to the method ch as a class member. We should try to avoid this. Looks to me like the best approach would be not to save the client as a class member and just build it using the builder when needed.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+            publishPkg();
+            return createBuild(artifactList, null);
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    /*
+     * Go pkg contains 3 files:
```
eyalbe4: Go pkg contains 3 files: --> The deployment of a go package requires 3 files:

(the go package is actually the zip).

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+     * */
+    private void publishPkg() throws IOException, NoSuchAlgorithmException {
+        createAndDeployZip();
+        deployGoMod();
+        createAndDeployInfo();
+    }
+
+    private void createAndDeployZip() throws IOException, NoSuchAlgorithmException {
+        // First, we create a temporary zip file of all project files.
+        String tmpZipPath = path.toString() + "/" + LOCAL_TMP_ZIP_FILENAME;
```
eyalbe4: Use File.Seperator here.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+        File tmpZipFile = archiveProjectDir(tmpZipPath);
+
+        // Second, filter the raw zip file according to Go rules and create deployable zip can be later resolved.
+        // We use the same code as Artifactory when he resolve a Go module directly from Github.
+        String deploayableZipPath = path.toString() + "/" + LOCAL_DEPLOYABLE_ZIP_FILENAME;
```
eyalbe4: Use File.Seperator here.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private void deployGoMod() throws IOException, NoSuchAlgorithmException {
+        String modLocalPath = getModFilePath();
+        Artifact deployedMod = deploy(new File(modLocalPath), PKG_MOD_FILE_EXTENSION);
+        artifactList.add(deployedMod);
+    }
+
+    private void createAndDeployInfo() throws IOException, NoSuchAlgorithmException {
+        String infoLocalPath = path.toString() + "/" + LOCAL_INFO_FILENAME;
```
eyalbe4: Use File.Seperator here.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+                Files.copy(filePath, zos);
+                zos.closeEntry();
+            }
+        }
+
+        return zipFile;
+    }
+
+    /*
+     * pkg info is a json file contains:
```
eyalbe4: ...json file contains: --> ...json file containing:

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,179 @@
...
+
+        infoMap.put("Version", version);
+        infoMap.put("Time", instant.toString());
+
+        mapper.writeValue(infoFile, infoMap);
+        return infoFile;
+    }
+
+    /* Deploy pkg file and add it as an buildInfo's artifact */
+    private Artifact deploy(File deployedFile, String extension) throws IOException, NoSuchAlgorithmException {
```
eyalbe4: No need to declare two exception types thrown by this method, because eventually, you are catching Exception.
You can therefore declare this method as throwing Exception and by that, also avoid re-declaring these two exception types by the upstream methods.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,220 @@
...
+            collectDependencies();
+            return createBuild(null, dependenciesList);
+        } catch (Exception e) {
+            logger.error(e.getMessage(), e);
+        }
+        return null;
+    }
+
+    /*
+     * In order to use Artifactory as a resolver we need to set GOPROXY env. var with Artifactory details.
```
eyalbe4: ...env. var... --> ...env var...

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,220 @@
...
+     * Wa also support fallback to VCS in case pkg doesn't exist in Artifactort,
+     */
+    private void setResolverAsGoProxy() throws IOException, URISyntaxException {
+        URL rtUrl = new URL(client.getArtifactoryUrl());
+        URIBuilder proxyUrlBuilder = new URIBuilder()
+                .setScheme(rtUrl.getProtocol())
+                .setUserInfo(resolverUsername, resolverPassword)
+                .setHost(rtUrl.getHost())
+                .setPath(rtUrl.getPath() + ARTIFACTORY_GO_API + resolutionRepository);
+        String proxyValue = proxyUrlBuilder.build().toURL().toString() + "," + GOPROXY_VCS_FALLBACK;
```
eyalbe4: Can this approach also be implemented in JFrog CLI? I think it is implemented differently there.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,220 @@
...
+    /*  Run 'go mod graph' and parse its output.
+     *  This command might change go.mod and go.sum content, so we backup and restore those files.
+     *    The output format is:
+     *     * For direct dependencies:
+     *          <module-name> <dependency's-module-name>@v<dependency-module-version>
+     *     * For transient dependencies:
+     *        <dependency's-module-name>@v<dependency-module-version> <dependency's-module-name>@v<dependency-module-version>
+     *   In order to populate build info dependencies, we parse the second column uf the mod graph output.
+     */
+    private void collectDependencies() throws IOException, InterruptedException, NoSuchAlgorithmException {
```
eyalbe4: We can probably declare only Exception as thrown here.

gradle.properties
```diff
@@ -1,8 +1,9 @@
...
+build-info-extractor-go-version=goL
+build-info-latest-release-version=2.13.L
+build-info-extractor-ivy-version=ivyL
 build-info-with-plugin=true
-build-info-version=2.13.x-SNAPSHOT
-artifactory-maven-plugin-version=2.7.x-SNAPSHOT
-build-info-extractor-maven3-version=2.13.x-SNAPSHOT
+build-info-version=2.13.L
+artifactory-maven-plugin-version=mvnL
+build-info-extractor-maven3-version=mvn3L
```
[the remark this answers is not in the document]
liron-shalom: Yes, I didn't mean to include this file on the PR.
For which initial version should I set the go-extractor?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
liron-shalom: Do you mean using "direct" in GOPROXY value?
If yes - the fallback approach is implemented differently in gocmd:
 First, we try to get the pkg from Artifactory and if we get an error we try to get it from the VCS server.
 I think we can change it and let the Go client do the fallback by itself.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,218 @@
...
+        env.put(GOPROXY_ENV_VAR, proxyValue);
+    }
+
+    private void runCmd() throws Exception {
+        List<String> args = new ArrayList<>();
+        for (String arg : goCmdArgs.split(" ")) {
+            args.add(arg);
+        }
+        runGoCmd(args);
+    }
```
yahavi: I think this function is redundant, because we can do its functionality in one line inside 'runGoGmd':
List<String> args = Lists.newArrayList(goCmdArgs.split(" "));

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,218 @@
...
+     *        <dependency's-module-name>@v<dependency-module-version> <dependency's-module-name>@v<dependency-module-version>
+     *   In order to populate build info dependencies, we parse the second column uf the mod graph output.
+     */
+    private void collectDependencies() throws Exception {
+        backupModAnsSumFiles();
+        List<String> goModArgs = new ArrayList<>();
+        goModArgs.addAll(Arrays.asList(GO_MOD_GRAPH_CMD.split(" ")));
+        CommandResults goGraphResult = runGoCmd(goModArgs);
+        String cachePath = getCachePath();
+        String dependenciesGraph[] = goGraphResult.getRes().split("\\r?\\n");
```
yahavi:
```suggestion
        String[] dependenciesGraph = goGraphResult.getRes().split("\\r?\\n");
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,218 @@
...
+     *    The output format is:
+     *     * For direct dependencies:
+     *          <module-name> <dependency's-module-name>@v<dependency-module-version>
+     *     * For transient dependencies:
+     *        <dependency's-module-name>@v<dependency-module-version> <dependency's-module-name>@v<dependency-module-version>
+     *   In order to populate build info dependencies, we parse the second column uf the mod graph output.
+     */
+    private void collectDependencies() throws Exception {
+        backupModAnsSumFiles();
+        List<String> goModArgs = new ArrayList<>();
```
yahavi:
```suggestion
        List<String> goModArgs = Lists.newArrayList(GO_MOD_GRAPH_CMD.split(" "));
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoPublish.java
```diff
@@ -0,0 +1,177 @@
...
+    private void createAndDeployZip(ArtifactoryBuildInfoClient client) throws Exception {
+        // First, we create a temporary zip file of all project files.
+        String tmpZipPath = path.toString() + File.separator + LOCAL_TMP_ZIP_FILENAME;
+        File tmpZipFile = archiveProjectDir(tmpZipPath);
+
+        // Second, filter the raw zip file according to Go rules and create deployable zip can be later resolved.
+        // We use the same code as Artifactory when he resolve a Go module directly from Github.
+        String deploayableZipPath = path.toString() + File.separator + LOCAL_DEPLOYABLE_ZIP_FILENAME;
+        File deployableZipFile = new File(deploayableZipPath);
+        GoZipBallStreamer pkgArchiver = new GoZipBallStreamer(new ZipFile(tmpZipFile), moduleName, version, logger);
```
yahavi:
```suggestion
        try (GoZipBallStreamer pkgArchiver = new GoZipBallStreamer(new ZipFile(tmpZipFile), moduleName, version, logger)) {
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoCommand.java
```diff
@@ -0,0 +1,107 @@
...
+            if (moduleNameRegexPattern.matcher(line).matches()) {
+                moduleName = line;
+            }
+        });
+        moduleName = StringUtils.split(moduleName, " ")[1];
+    }
+
+    protected void preparePrerequisites(String repo, ArtifactoryBaseClient client) throws VersionException, IOException {
+        validateArtifactoryVersion(client);
+        validateRepoExists(repo, client,"Target repo must be specified");
```
yahavi: Shouldn't it be "Source repo must be specified" when running go-build?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
```diff
@@ -0,0 +1,218 @@
...
+                    .md5(checksums.get(MD5)).sha1(checksums.get(SHA1))
+                    .type("zip")
+                    .build();
+            dependenciesList.add(dependency);
+        }
+    }
+
+    /*
+     * Run go client cmd with goArs.
+     * Write stdout + stderr to looger, and return the command's result.
```
yahavi:
```suggestion
     * Write stdout + stderr to logger, and return the command's result.
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
liron-shalom: Eyal told me we should try to avoid using guava as much as possible in build-info code. Is there other one-line we can use?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
liron-shalom: Same as before - Can we avoid using guava?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoRun.java
yahavi: List<String> args = new ArrayList<>(Arrays.asList(goCmdArgs.split(" ")));

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoCommand.java
liron-shalom: The repo existence is checked in run and publish. I changed the error msg to be more general.

PR #287, opened 2019-12-06
Reintroduce buildinfo computation parallelism

ghale:
This supercedes #275 and fixes the issues reported there.  I've tested this with Gradle 4.10.3, 5.6.4, and 6.0 using the example project [here](https://github.com/jfrog/project-examples/blob/master/gradle-examples/gradle-example-publish/build.gradle).

I'm happy to discuss if any of the changes are unclear.

PR #301, opened 2020-01-23
Improve module info extraction to gather info safely in parallel (only)

ghale:
This is a followup from #293.  Given the difficulties with incorporating the integration tests I included there, I'm carving off the functional change that I made in that PR so that it can be tested as normal and we can resolve the issues with the integration tests I added as a separate PR.

This refactors the module info extraction to make it safer. With #287 this was parallelized in a fairly unsafe manner using some Gradle internal APIs. This breaks up the module info extraction so that instead of a single task in the root project reaching across and resolving configurations in other projects, each project has its own extraction task that produces a module info artifact by resolving only the configurations in its project. Then the root project has a configuration that aggregates all of the module info artifacts and adds them to the build info object. Since this is done in a more "standard" way from a Gradle perspective, it allows us to remove the internal Gradle API usages. Since those module info tasks run in different projects, they can safely execute in parallel prior to the deploy task in the root project starting. This allows us to push some of this work earlier in the build and run it in parallel to other stuff that might be blocking the main deploy task. YMMV, but in certain cases this could provide a fair performance improvement.

We've learned that some of the unsafe things we did in #287 can lead to a deadlock condition in some scenarios.  The changes in this PR should avoid the potential deadlock.

cc @ljacomet

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
@@ -48,13 +48,16 @@ public class GradleDeployDetails implements Comparable<GradleDeployDetails> {
     public PublishArtifactInfo getPublishArtifact() {
         return publishArtifact;
     }
-    
+
     public int compareTo(GradleDeployDetails that) {
         if (this.publishArtifact == null) {
             return -1;
         }
         if (that.publishArtifact == null) {
             return 1;
+        }
+        if (this.deployDetails.compareTo(that.deployDetails) == 0) {
+            return 0;
         }
         String thisExtension = this.publishArtifact.getExtension();
         String thatExtension = that.publishArtifact.getExtension();
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
@@ -56,9 +56,12 @@ public class GradleDeployDetails implements Comparable<GradleDeployDetails> {
         if (that.publishArtifact == null) {
             return 1;
         }
-        if (this.deployDetails.compareTo(that.deployDetails) == 0) {
+
+        int compareDeployDetails = this.deployDetails.compareTo(that.deployDetails);
+        if (compareDeployDetails == 0) {
             return 0;
         }
+
         String thisExtension = this.publishArtifact.getExtension();
         String thatExtension = that.publishArtifact.getExtension();
         if (thisExtension == null) {
@@ -75,7 +78,7 @@ public class GradleDeployDetails implements Comparable<GradleDeployDetails> {
         if ("xml".equals(thatExtension) || "pom".equals(thatExtension)) {
             return -1;
         }
-        return this.deployDetails.compareTo(that.deployDetails);
+        return compareDeployDetails;
     }
 
     @Override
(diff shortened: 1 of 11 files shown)


ghale:
Thanks @yahavi, that should be fixed.  

I was only testing against the `gradle-example-publish` build and the problem was triggered by `publishConfigs()` which isn't in that build (it is in `gradle-example` which is why that build triggered the failure).  What other builds in there should I try in order to get good coverage for a PR - should I try all of them?

yahavi:
@ghale, I think that gradle-example and gradle-example-publish is enough.
I tried it and the last commit did solve the issue I mentioned. However I have noticed a strange behaviour: The pom files are published twice.

Steps to reproduce:

1. Install build-info dependencies: `gradle clean install`
2. Clone https://github.com/jfrog/project-examples.git
3. cd `gradle-examples/gradle-example-minimal/`
4. Configure the buildscript in `build.gradle` to use the installed build-info dependencies:
5. Run `gradle clean artifactoryPublish`

The results:
```
Deploying artifact: http://127.0.0.1:8081/artifactory/libs-snapshot-local/gradle-example-minimal/1.0-SNAPSHOT/gradle-example-minimal-1.0-SNAPSHOT.jar
Deploying artifact: http://127.0.0.1:8081/artifactory/libs-snapshot-local/gradle-example-minimal/1.0-SNAPSHOT/gradle-example-minimal-1.0-SNAPSHOT.pom
Deploying artifact: http://127.0.0.1:8081/artifactory/libs-snapshot-local/gradle-example-minimal/1.0-SNAPSHOT/gradle-example-minimal-1.0-SNAPSHOT.pom
```

ghale:
@yahavi I've fixed that, but I don't understand enough about the way `ArtifactoryTask` gathers artifacts.  The issue is that `ArtifactoryTask.collectDescriptorsAndArtifactsForUpload()` now gets called twice - once when we gather the module info, and once when we aggregate them all together.  Because `ArtifactoryTask.deployDetails` is a set, I assumed that this would be idempotent and would only add each artifact once.  However, looking at `GradleDeployDetails` the rules in `compareTo()` are a little more complex.  Specifically, if the artifact is a pom or ivy xml file, then it always treats it as different and adds it to the set (see https://github.com/jfrog/build-info/blob/efcdd4c73f7930d06934730351ae49c8d9e333b3/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java#L67-L74).  It's not clear to me why we wouldn't still check the path to the artifact, but I didn't want to change it without understanding it.

The change I made was to only execute the body of `collectDescriptorsAndArtifactsForUpload()` if the set is already empty, which means we only attempt to add things to the set once if it gets called multiple times.  If there's a better way to deal with this, let me know.

yahavi:
@ghale I think the reason for the special treatment for poms is to deploy them in the end. I guess that until now we called `collectDescriptorsAndArtifactsForUpload()` only once, so the issue didn't show up. 
Is it necessary to call `collectDescriptorsAndArtifactsForUpload()` twice? If not, lets call it once.
Otherwise, I'd suggest to return 0 before `thisExtension = thisExtension.toLowerCase();` if `this.deployDetails.compareTo(that.deployDetails) == 0`.
Thanks!

ghale:
Thanks @yahavi.  I ended up doing both - we only call `collectDescriptorsAndArtifactsForUpload()` once and I also fixed the `compareTo()` method for `GradleDeployDetails` such that it will not add duplicates but will still sort the pom/ivy xmls as last in the set.  That should clear things up.

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
```diff
@@ -48,14 +48,17 @@ public Project getProject() {
...
-    
+
     public int compareTo(GradleDeployDetails that) {
         if (this.publishArtifact == null) {
             return -1;
         }
         if (that.publishArtifact == null) {
             return 1;
         }
+        if (this.deployDetails.compareTo(that.deployDetails) == 0) {
```
yahavi: `compareTo` triggers a big string comparison. Can you please save the result in a variable and reuse it in the last line?

yahavi:
@ghale, Thanks! Looks like issue is resolved! 
One last thing - I wrote a comment in [GradleDeployDetails:59](https://github.com/jfrog/build-info/pull/301/files#diff-713919fef5d97efc6eae875736466e6bR59). Fix that and I'll merge.

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleDeployDetails.java
ghale: Done

PR #306, opened 2020-02-04
Mvn3Extractor - Support insecure tls for maven resolution and deployment.

liron-shalom:
*This commit includes refactor PreemptiveHttpClient to be created by a builder.


build-info-client/src/main/java/org/jfrog/build/client/PreemptiveHttpClientBuilder.java
```diff
@@ -0,0 +1,228 @@
...
+
+        HttpClientBuilder builder = HttpClientBuilder
+                .create()
+                .setConnectionManager(connectionManager)
+                .setDefaultRequestConfig(requestConfig);
+
+        // Add as the first request interceptor
+        builder.addInterceptorFirst(new PreemptiveHttpClient.PreemptiveAuth());
+
+        // set the following user agent with each request
```
eyalbe4: // set... --> // Set...

PR #310, opened 2020-02-17
Add support for Customize module name

liron-shalom:


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/ArtifactoryClientConfiguration.java
```diff
@@ -504,6 +504,16 @@ public String getNpmPath() {
...
+
+        public String getNpmModule() {
+            return rootConfig.getStringValue(NPM_MODULE);
+        }
+
+        public void setNpmModule(String npmModule) {
+            rootConfig.setStringValue(NPM_MODULE, npmModule);
+        }
+
+
```
eyalbe4: Redundant new-lines.

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/ClientConfigurationFields.java
```diff
@@ -33,6 +33,7 @@
     String NPM_INSTALL_ARGS = "npm.installArgs";
     String NPM_EXECUTABLE_PATH = "npm.executablePath"; // Path to npm CLI
     String NPM_PATH = "npm.path"; // Path to package.json dir
+    String NPM_MODULE = "npm.module"; // custom module name for npm operation's build-info
```
eyalbe4: custom module name for npm operation's build-info --> Custom npm module name for the build-info

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoCommand.java
```diff
@@ -41,12 +41,14 @@
     ArtifactoryClientBuilderBase clientBuilder;
     Path path;
     String moduleName;
+    String buildInfoModuleId;
```
eyalbe4: Question - Why do we need both the *moduleName* and the *buildInfoModuleId* fields? I can see that *buildInfoModuleId* is the customised one, but can't we store the customised one in *moduleName*? Why do we need to store both?
Also related to this - it looks lie we call *parseModuleName()* even if we get a customised module name. Is this necessary?

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
```diff
@@ -66,7 +68,8 @@ public Build extract(NpmProject npmProject) throws Exception {
         restoreNpmrc(workingDir);
 
         List<Dependency> dependencies = collectDependencies(workingDir);
-        return createBuild(dependencies);
+        String moduleId = org.apache.commons.lang.StringUtils.isNotBlank(module) ? module : npmPackageInfo.toString();
```
eyalbe4: Any reason why we use **org.apache.commons.lang.StringUtils** and not the imported **StringUtils**?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoCommand.java
liron-shalom: "module name" is go figure which is the unique ID of the module in the "go world", we are specify it in the go.mod, and use it to publish and import.
We cannot force the buildInfoModuleId in case the user want a different one from the name was specified in the project. (For example if he wants to aggregate several pkgs under one module, it will be impossible to import them and use them as dependencies.)

PR #318, opened 2020-03-06
Npm Extractor - allow JSON output only if it was required

liron-shalom:


build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
```diff
@@ -159,6 +162,10 @@ private void createTempNpmrc(Path workingDir, List<String> installationArgs) thr
         }
     }
 
+    private boolean isJsonOutputRequiered(List <String> installationArgs) {
+        return installationArgs.contains("--json") || installationArgs.contains("--json=true");
```
yahavi: 1. `installationArgs.contains("--json=true")` infer `installationArgs.contains("--json")`. `installationArgs.contains("--json")` is enough.
2. This will not work on 
```
--json=false
--json false
--json        false
--json true
--json        true
```

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
```diff
@@ -159,6 +162,20 @@ private void createTempNpmrc(Path workingDir, List<String> installationArgs) thr
...
+     *  1. --arg - which infers true
+     *  2. --arg=value (true/false)
+     *  3. --arg value (true/false)
+     */
+    private boolean isJsonOutputRequired(List <String> installationArgs) {
+        int jsonIndex = installationArgs.indexOf("--json");
+        if (jsonIndex > -1 ) {
+            return jsonIndex == installationArgs.size() - 1 || !installationArgs.get(jsonIndex).equals("false");
+        }
+        return installationArgs.contains("--json=true");
```
yahavi: If `jsonIndex == -1`, we don't have `--json` in `installationArgs`.
Conclusion: `installationArgs.contains("--json=true")` is always false.

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
```diff
@@ -159,6 +162,20 @@ private void createTempNpmrc(Path workingDir, List<String> installationArgs) thr
...
+    /**
+     * Boolean argument can be provided in one of the following ways:
+     *  1. --arg - which infers true
+     *  2. --arg=value (true/false)
+     *  3. --arg value (true/false)
+     */
+    private boolean isJsonOutputRequired(List <String> installationArgs) {
+        int jsonIndex = installationArgs.indexOf("--json");
+        if (jsonIndex > -1 ) {
+            return jsonIndex == installationArgs.size() - 1 || !installationArgs.get(jsonIndex).equals("false");
```
yahavi: I wonder if this code would return false if we have`--json=false`. 🤔 
I think that this function deserved a unit test.

liron-shalom:
We split the raw string args into a list by whitespace:
` this.installArgs = StringUtils.isBlank(installArgs) ? new ArrayList<>() : Arrays.asList(installArgs.trim().split("\\s+"));`
so the case of "--arg=value" is treated as one element in the list/

Issue #325, opened 2020-03-12
Gradle plugin: ExtractModuleTask emits warnings on every build using Gradle 6.3+


rpalcolea:
This warning started showing up with Gradle 6.3:

```
java.lang.IllegalArgumentException: Deprecation warnings were found (Set the ignoreDeprecations system property during the test to ignore):
 - Type 'ExtractModuleTask': field 'moduleFile' without corresponding getter has been annotated with @OutputFile. This behaviour has been deprecated and is scheduled to be removed in Gradle 7.0. See https://docs.gradle.org/6.3-rc-1/userguide/more_about_tasks.html#sec:up_to_date_checks for more details.
``` 

This is because it is a java class and needs a getter instead of annotating the property only. This wouldn't be the case for a groovy one.

As a side note, I think it would be great to migrate the build to modern gradle so this issues even prevent this project from building because the validation of the plugin project would fail.

Current state, if you run `./gradlew :build-info-extractor-gradle:validateTaskProperties`  you will see

```
> Task :build-info-extractor-gradle:validateTaskProperties
Task property validation finished with warnings:
  - Warning: Type 'org.jfrog.gradle.plugin.artifactory.task.ExtractModuleTask': field 'moduleFile' without corresponding getter has been annotated with @OutputFile
```

Issue #302, opened 2020-01-24
Add integration tests for Gradle


ghale:
#293 introduced some rudimentary integration tests for Gradle.

comments from @yahavi 

> We also had a discussion in my team about the tests.
> As you see, currently we don't have integration tests on build-info-extractor-gradle and we are undoubtedly need them.
> I know you worked very hard on them, but there are issues we have to deal with, before merging the code. Please let me know what you think:
> 
> We use TestNG in our integration tests. From our side, It is not very clever to use 2 test environments. Please switch Spoke to TestNG.
> We run integration tests once for all modules. The integration tests shouldn't download Artifactory OSS. The reason is - In other modules we test many of Artifactory features, including once that doesn't enabled in the OSS. It also take a lot of time to download. The solution is to provide an existing Artifactory details. You can download and start a local Artifactory OSS in you machine and provide its url, username and password to the test as environment variables.
> The good news are - We already have tests infrastructure for all of the above! The integration tests should extend IntegrationTestsBase. In this base class we create and delete local remote and virtual repositories in Artifactory. Additionally, we use this class to create the Artifactory clients.
> If you need a good reference to write integration tests for modules see GoExtractorTest and NpmExtractorTest.
> For more information see also: Building and Testing the Sources.

Comments from @ghale 

> I tried to implement my test using IntegrationTestsBase per your recommendation. The problem is that this class tries to create repositories in Artifactory via the rest API and according to the docs that function requires Artifactory Pro. Running the test against Artifactory OSS gives me an error on that step.
> 
> The fundamental problem here is that it's difficult to contribute anything meaningful to this project without some sort of automated regression testing. In fact, my primary motivation for adding the automated testing in the first place was because I wasn't comfortable doing the refactor in this PR without some regression tests in the first place.
> 
> To use theIntegrationTestsBase framework, it seems like you 1) have to manually download, start and manage an Artifactory instance and 2) that instance needs to be Artifactory Pro. To use the Pro version, you need a trial license key which then means you'll be contacted (repeatedly) by the sales team, which is a strong disincentive for contributing. If you're not working on this project on a regular basis, the chances are you'll have to do this every time.
> 
> Ultimately, I think it would be great if there were an automated way to bootstrap the whole download/start/configure step, but even if it were manual, as long as it's painless, I think it's a step in the right direction. Based on my current understanding, I'm not sure I see a clear path to this with how things are now.
> 
> Some questions related to this:
> 
[...]

PR #326, opened 2020-03-12
ExtractModuleTask should have a proper getter for moduleFile property

rpalcolea:
This fixes https://github.com/jfrog/build-info/issues/325

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/ExtractModuleTask.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/ExtractModuleTask.java
@@ -13,7 +13,7 @@ import java.lang.reflect.InvocationTargetException;
 
 public class ExtractModuleTask extends DefaultTask {
 
-    public RegularFileProperty moduleFile;
+    private RegularFileProperty moduleFile;
 
     public ExtractModuleTask() {
         try {
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/ExtractModuleTask.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/ExtractModuleTask.java
@@ -12,7 +12,7 @@ import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
 
 public class ExtractModuleTask extends DefaultTask {
-    @OutputFile
+
     public RegularFileProperty moduleFile;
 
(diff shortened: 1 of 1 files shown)


rpalcolea:
Hi @yahavi , any chance you could take a look at this one?

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/ExtractModuleTask.java
```diff
@@ -12,7 +12,7 @@
 import java.lang.reflect.InvocationTargetException;
 
 public class ExtractModuleTask extends DefaultTask {
-    @OutputFile
+
     public RegularFileProperty moduleFile;
```
wolfs: Can you make the field `private` and `final`? Or would that be a breaking change?

rpalcolea: I changed it to `private`. However, changing it to final would require making a change in the constructor:

```
   public ExtractModuleTask() {
        try {
            this.moduleFile = getProject().getObjects().fileProperty();
        } catch (NoSuchMethodError e) {
            // Gradle 4.x
            this.moduleFile = invokeMethod(getProject().getLayout(), "fileProperty");
        }
    }
```

Since `moduleFile` could potentially be assigned twice so java compiler doesn't like it

This leaves me again in https://github.com/jfrog/build-info/issues/302 where a better story around testability of this plugin would help to make changes with more confident. I think having it as `private` to start is a move in the right direction

rpalcolea: Actually, I think this is better: https://github.com/jfrog/build-info/pull/326/commits/0b8a578296eca1cdaaff61830029b59ddc880e42  instead of relying on the try/catch

eyalbe4:
@rpalcolea,
Thanks for a lot for this fix.
It looks though that you haven't joined our contributors community by signing [JFrog's CLA](https://secure.echosign.com/public/hostedForm?formid=5IYKLZ2RXB543N). Can you sign it please?

eyalbe4:
@rpalcolea - disregard my last comment.
I see you signed the CLA already.
Thanks!

rpalcolea:
Hi @eyalbe4 @yahavi 
Thanks again for looking at this

Are there plans to release this soon?

yahavi:
@rpalcolea 
Gradle Artifactory plugin 4.15.2 is released.

PR #336, opened 2020-04-19
Wire module info files without using dependencies

### LINKED ISSUE #334: Gradle plugin does not publish all artifacts
Using same module name as the root project causes the extractModuleInfo task of the module to not being triggered.
Reproducer with instructions here - https://github.com/yahavi/gradle-deploying-issue

It looks like [builtBy](https://github.com/jfrog/build-info/pull/301/files#diff-b2a12c1b329a701fe714eb55df17d8edR140) does not happen when it already been called once for a project with the same GAV.

@ghale can you please take a look? Thanks!

ghale:
#334 describes a problem whereby when the root project does not produce publishable artifacts, but has the same coordinates as a subproject, the subproject will not be published.  This happens because the inter-project dependencies are wired together via artifacts and when two projects have matching coordinates, they appear as the same module so conflict resolution kicks in and resolves to one project or the other.  If this were two subprojects, there would be no reason to apply the artifactory plugin to both projects (otherwise they would both publish the same module).  But in the case where the root project does not publish any artifacts, it is entirely feasible to apply the plugin to both the root project and the subproject event though the coordinates are the same.

These changes fix the issue by wiring the module info files to the deploy task without using dependencies.  This is generally undesirable because we would like to have all inter-project dependencies modeled as project dependencies, but Gradle currently can't handle the peculiarities of this use case.  Once this is fixed in Gradle, we can change this back to using dependency-based relationship between projects.

Fixes #334

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPluginBase.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPluginBase.groovy
@@ -16,21 +16,19 @@
 
 package org.jfrog.gradle.plugin.artifactory
 
-import org.gradle.api.Named
 import org.gradle.api.Plugin
 import org.gradle.api.Project
-import org.gradle.api.artifacts.Configuration
-import org.gradle.api.attributes.Attribute
+import org.gradle.api.file.FileCollection
 import org.jfrog.gradle.plugin.artifactory.dsl.ArtifactoryPluginConvention
 import org.jfrog.gradle.plugin.artifactory.extractor.listener.ProjectsEvaluatedBuildListener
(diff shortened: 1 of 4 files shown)


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/ArtifactoryPluginBase.groovy
```diff
@@ -165,14 +138,31 @@ abstract class ArtifactoryPluginBase implements Plugin<Project> {
...
+            this.extractModuleTask = extractModuleTask
+        }
+
+        @Override
+        boolean hasModules() {
+            if (artifactoryTask.project.getState().getExecuted()) {
+                return artifactoryTask != null && artifactoryTask.hasModules();
+            } else {
+                return false;
+            }
```
yahavi, line 160: Remove the 'else'
```suggestion
            }
            return false;
```

Issue #342, opened 2020-05-08
Gradle Build Info Extractor should not add modules without artifacts


rpalcolea:
Hi folks,

We noticed that in previous versions of the plugin did not add modules to the build info if the modules did not have artifacts. These are the original pieces of code:

```
 Set<Project> allProjects = rootProject.getAllprojects();
        for (Project project : allProjects) {
            if(project.getState().getExecuted()) {
                ArtifactoryTask buildInfoTask = getBuildInfoTask(project);
                if (buildInfoTask != null && buildInfoTask.hasModules()) {
                    bib.addModule(extractModule(project));
                }
            }
        }

public boolean hasModules() {
        return helperConfigurations.hasModules() || helperPublications.hasModules();
    }
```

When a project didn't have modules, it wouldn't add it

With the current implementation:  https://github.com/jfrog/build-info/blob/master/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java#L108-L114 this not the case anymore

Now it only reads the module information and adds it to the list of modules to publish as part of build info

However, if a module looks like this:

```
{
  "id" : "test.nebula.netflix:sub2:1588981693307",
  "artifacts" : [ ],
  "excludedArtifacts" : [ ],
  "dependencies" : [ ]
}

``` 

The end result in artifactory is:

![image](https://user-images.githubusercontent.com/1625920/81457665-e93ec080-914b-11ea-8c6b-6cf1cf9cf323.png)

I think this should be prevented to avoid bad metadata in the build information

PR #343, opened 2020-05-08
GradleBuildInfoExtractor: only add modules with artifacts to build info

rpalcolea:
This addresses "Gradle Build Info Extractor should not add modules without artifacts" https://github.com/jfrog/build-info/issues/342


eyalbe4:
@rpalcolea,
Thanks for creating this PR!
If a module does not include artifacts, but does include depedencies, it should still be added to the build-info. Otherwise, the info about the dependencies will not be recorded.
I guess that we need check whether a module does not include artifacts and depdencies.

rpalcolea:
Hi @eyalbe4 

Thanks for looking at this 

I did the change =)

PR #351, opened 2020-06-29
Default MavenJava and IvyJava publications added in non Java projects

yahavi:
-----

**The issue**
In CI builds, MavenJava and IvyJava default publications are automatically injected. This behavior causes issue when running, for example, C++ project with Gradle.

**Solution**
Inject MavenJava and IvyJava only if `java` or `java-library` plugin are applied.

**More information**
Jenkins Jira isssue: https://www.jfrog.com/jira/browse/HAP-1307

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
@@ -109,7 +109,7 @@ public class ProjectsEvaluatedBuildListener extends BuildAdapter implements Proj
             if (resolver != null) {
                 defineResolvers(artifactoryTask.project, resolver)
             }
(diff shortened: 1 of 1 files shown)


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
```diff
@@ -108,13 +109,27 @@ public class ProjectsEvaluatedBuildListener extends BuildAdapter implements Proj
...
             artifactoryTask.projectEvaluated()
         }
     }
 
+    /**
+     * Return true if the 'java' or 'java-library' plugins are applied.
+     * @param artifactoryTask - The artifactory task
+     * @return true if the 'java' or 'java-library' plugins are applied.
+     */
+    private static boolean isJava(ArtifactoryTask artifactoryTask) {
```
eyalbe4, line 124: I think it makes more sense of this method ro receive a *Project* as an argument, rather than a task.

eyalbe4:
https://www.jfrog.com/jira/browse/GAP-318 tracks this change.

PR #368, opened 2020-08-06
Changed jdk and credentials

asaf12355:
@eyalbe4 please review

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -70,7 +70,7 @@ node('java') {
 
     // Prepare build env.
     cleanWs()
-    def jdktool = tool name: "1.8.0_131"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -135,18 +133,14 @@ node('java') {
                     // Create tag.
                     def tag = buildConfig.tagName + "-" + buildConfig.releaseVersion
                     sh "git tag $tag"
-                    wrap([$class: 'MaskPasswordsBuildWrapper', varPasswordPairs: [[password: 'GITHUB_API_KEY', var: 'SECRET']]]) {
```
eyalbe4, line 138: I suspect that removing will cause the password to be included in the build log.
Let me know if you tested this and the key is masked in the log.

PR #386, opened 2020-09-30
Set custom java temp dir for docker java.

Or-Gabay:
-----
Fixes https://github.com/jfrog/jenkins-artifactory-plugin/issues/268#issuecomment-678318546

### DIFF
--- a/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerUtils.java
+++ b/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerUtils.java
@@ -229,7 +229,7 @@ public class DockerUtils {
         return indexOfFirstSlash < indexOfLastColon;
     }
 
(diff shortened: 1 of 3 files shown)


build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerUtils.java
```diff
@@ -226,4 +228,11 @@ public static Boolean isImageVersioned(String imageTag) {
         int indexOfLastColon = imageTag.lastIndexOf(":");
         return indexOfFirstSlash < indexOfLastColon;
     }
+
+    // Set the Java temp dir system property. As a result, java will create it for us.
```
eyalbe4, line 232: Let's add to this description the reason for why we're doing this. Without it, the random reader will have no idea why this is done.

eyalbe4 (changes requested):

PR #405, opened 2020-10-28
BI-540 - The Xray client does not support access tokens

yahavi:
-----

The username-password constructor left intact in order to not break the API.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryXrayClient.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryXrayClient.java
@@ -44,9 +44,13 @@ public class ArtifactoryXrayClient extends ArtifactoryBaseClient {
      */
     private static final int HTTP_CLIENT_RETRIES = 0;
 
-    public ArtifactoryXrayClient(String artifactoryUrl, String username, String password, Log logger) {
-        super(artifactoryUrl, username, password, StringUtils.EMPTY, logger);
+    public ArtifactoryXrayClient(String artifactoryUrl, String username, String password, String accessToken, Log logger) {
+        super(artifactoryUrl, username, password, accessToken, logger);
         setConnectionRetries(HTTP_CLIENT_RETRIES);
+    }
+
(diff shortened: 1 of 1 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryXrayClient.java
```diff
@@ -44,11 +44,15 @@
      */
     private static final int HTTP_CLIENT_RETRIES = 0;
 
-    public ArtifactoryXrayClient(String artifactoryUrl, String username, String password, Log logger) {
-        super(artifactoryUrl, username, password, StringUtils.EMPTY, logger);
+    public ArtifactoryXrayClient(String artifactoryUrl, String username, String password, String accessToken, Log logger) {
```
eyalbe4, line 47: Don’t we need a constructor that received only access token? Why do we need to pass username, password and accessToken?

eyalbe4 (changes requested):

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryXrayClient.java
yahavi: This Base client contains the following logic:
```java
if (StringUtils.isNotEmpty(accessToken)) {
    httpClient = new ArtifactoryHttpClient(this.artifactoryUrl, accessToken, logger);
} else {
    httpClient = new ArtifactoryHttpClient(this.artifactoryUrl, username, password, logger);
}
```
With that, we don't have to add the above check in the Jenkins plugin. For example, in the Jenkins plugin I did this:

```java
ArtifactoryXrayClient client = new ArtifactoryXrayClient(server.getUrl(), credentials.getUsername(), credentials.getPassword(), credentials.getAccessToken(), log);

```

eyalbe4: Got it. OK.

eyalbe4 (approved):

Issue #406, opened 2020-10-30
Android .aar Artifacts not building


sebasira:
**Describe the bug**
Artifacts are not generated for Android libraries (bu they are for other Java/gradle base modules) from GitLab Ci/CD YAML script using jFrog CLI. 
Yesterday I notice that a new version of an Android library was not deployed to Artifactory, at first it seems strange because GitLab deploy job succeed, but looking at the logs everything went fine but it seems like some gradle task were skipped. Please look at the logs I posted instead of the screenshot.

**To Reproduce**
Here's part of the GitLab YAML script I'm using:

```
.install_gradle: &install_gradle |
  curl -L https://services.gradle.org/distributions/gradle-$GRADLE_VERSION-all.zip -o gradle.zip
  unzip -q gradle.zip
  export GRADLE_HOME=$(pwd)/gradle-$GRADLE_VERSION
  export PATH=${GRADLE_HOME}/bin:${PATH}
  echo $GRADLE_HOME
  echo $PATH
  gradle -v

.connect_artifactory: &connect_artifactory |
  curl -fL https://getcli.jfrog.io | sh -s v1.39.7
  ./jfrog rt config --url=$ARTIFACTORY_URL --user=$ARTIFACTORY_USER --password=$ARTIFACTORY_PASS
  ./jfrog rt c show


.upload_artifact: &upload_artifact |
  ./jfrog rt gradle-config --use-wrapper=true --repo-resolve=$GRADLE_REPO_KEY --server-id-resolve=Default-Server --repo-deploy=$GRADLE_REPO_KEY --server-id-deploy=Default-Server
  ./jfrog rt gradle clean artifactoryPublish -b build.gradle --build-name=$ARTIFACT_NAME --build-number=$CI_JOB_ID


# JOB:
publish:
  stage: deploy
  cache: {}
  script:
    # Download and Install Gradle
    - *install_gradle
    # Connect to JFrog Artifactory
    - *connect_artifactory
    # Replace the repository name in the configuration.yml to the correct one.
    - sed -i 's,GRADLE_REPO_KEY,'"$GRADLE_REPO_KEY"',g' configuration.yaml
    # Upload artifact to JFrog Artifactory
    - *upload_artifact
```

This script is the same for Android and Java (Spring Boot) gradle based modules. For Java modules the artifact are generated and deployed, but they are not for Android.

**Expected behavior**
What I would expect is first that the Gradle task that generates the artifact were executed and then that artifact deployed to Artifactory.

[...]

PR #407, opened 2020-10-31
Gradle CI - Add existing publications by default

yahavi:
-----

Gradle users who use our plugins (CLI, Jenkins, Bamboo, etc...) to publish Gradle Artifacts, typically shouldn't add any Artifactory Gradle plugin configuration to their build.gradle.
Currently, for Java users in Gradle CI mode, we injects automatically the MavenJava and IvyJava publications. This is good to publish Jars. However, sometimes this behavior is not enough, since publications configured in `build.gradle` should also be also included. 

From the suggested new logic, I'm excluding users who did specified their publications. For example:
```groovy
artifactoryPublish {
    publications(publishing.publications.aar)
}
```

This PR is possibly resolving the following issue:
https://github.com/jfrog/build-info/issues/406, https://github.com/jfrog/jfrog-cli/issues/880
https://stackoverflow.com/questions/64606969/jfrog-cli-and-gradle-artifacrotypublish-not-generating-android-artifacts

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
@@ -11,6 +11,7 @@ import org.gradle.api.artifacts.repositories.IvyArtifactRepository
 import org.gradle.api.component.SoftwareComponent
 import org.gradle.api.invocation.Gradle
 import org.gradle.api.publish.Publication
+import org.gradle.api.publish.PublicationContainer
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.ivy.IvyPublication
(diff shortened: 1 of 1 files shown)


eyalbe4:
Update - 
Following further discussions about the preferred approach, we think it may be best to implement the following instead:
1. If no publications are specified, the plugin will deploy no artifacts (same behaviour as today).
2. If one or more publications are specified, they should be used (same behaviour as today).
3. We will add a new ALL argument, which the *publications* method can accept, which will trigger the usage of all publications configured in the build script.
4. When used from a CI server, you'll have the option of specifying which publications you'd like to use (Jenkins already implemented this - see [here](https://www.jfrog.com/confluence/display/JFROG/Declarative+Pipeline+Syntax#DeclarativePipelineSyntax-GradleBuildswithArtifactory) and [here](https://www.jfrog.com/confluence/display/JFROG/Scripted+Pipeline+Syntax#ScriptedPipelineSyntax-GradleBuildswithArtifactory). This should still be implemented in JFrog CLI, as part of the **jfrog rt gradle-config** command. Same goes for the other CI integrations.
5. All CI integrations will support the ALL option described in section 3 above.

PR #409, opened 2020-11-11
Bugfix - docker push fail to collect build info while using repo inst…

Or-Gabay:
…ead subdomain for docker registry

-----

### DIFF
--- a/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/types/DockerImage.java
+++ b/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/types/DockerImage.java
@@ -262,13 +262,13 @@ public class DockerImage implements Serializable {
         String ImagePath = DockerUtils.getImagePath(imageTag);
         ArrayList<String> manifestPathCandidate = new ArrayList<>(DockerUtils.getArtManifestPath(ImagePath, targetRepo, cmdType));
         logger.info("Trying to fetch manifest from Artifactory.");
-        int listLen = manifestPathCandidate.toArray().length;
+        int listLen = manifestPathCandidate.size();
         for (int i = 0; i < listLen; i++) {
             try {
                 checkAndSetManifestAndImagePathCandidates(manifestPathCandidate.get(i), dependenciesClient);
                 return;
             } catch (IOException e) {
-                // Ignore - unless we have no search path.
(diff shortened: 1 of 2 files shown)


build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/types/DockerImage.java
```diff
@@ -257,41 +259,20 @@ private DockerLayers getLayers(ArtifactoryDependenciesClient dependenciesClient,
...
-            logger.error("The manifest could not be fetched from Artifactory - " + e.getMessage());
-            // If image path includes more than 3 slashes, Artifactory doesn't store this image under 'library',
-            // thus we should not look further.
-            int totalSlash = StringUtils.countMatches(proxyImagePath, "/");
-            if (cmdType == DockerUtils.CommandType.Push || totalSlash > 3) {
-                throw e;
+        String ImagePath = DockerUtils.getImagePath(imageTag);
+        ArrayList<String> manifestPathCandidate = new ArrayList<>(DockerUtils.getArtManifestPath(ImagePath, targetRepo, cmdType));
+        logger.info("Trying to fetch manifest from Artifactory.");
+        int listLen = manifestPathCandidate.toArray().length;
```
eyalbe4, line 265: Let's replace the above line with:
```
int listLen = manifestPathCandidate.size();
```

build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/types/DockerImage.java
```diff
@@ -257,41 +259,20 @@ private DockerLayers getLayers(ArtifactoryDependenciesClient dependenciesClient,
...
+        for (int i = 0; i < listLen; i++) {
+            try {
+                checkAndSetManifestAndImagePathCandidates(manifestPathCandidate.get(i), dependenciesClient);
+                return;
+            } catch (IOException e) {
+                // Ignore - unless we have no search path.
```
eyalbe4, line 271: I suggest changing this comment to the following, but please see if this is what you actually wanted to say here:
// Throw the exception only if we reached the end of the loop, which means we tried all options.

eyalbe4 (changes requested):

PR #418, opened 2020-11-30
Exception raised during download when 'java.io.tmpdir' is not defined

yahavi:
-----


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/spec/SpecsHelper.java
```diff
@@ -133,6 +134,7 @@ public SpecsHelper(Log log) {
      * @throws IOException in case of IOException
      */
     public List<Dependency> downloadArtifactsBySpec(String spec, ArtifactoryDependenciesClient client, String targetDirectory) throws IOException {
+        handleJavaTmpdirProperty();
```
eyalbe4, line 137: Let's explain here why it is important to handle the temp dir as part of the download flow.
Let's add a similar explanation in the ```PackageManagerExtractor``` class. I think this will rally help the random reader.

eyalbe4 (changes requested):

PR #426, opened 2020-12-22
Tests - fix possible NPE

RobiNino:


build-info-extractor/src/test/java/org/jfrog/build/extractor/clientConfiguration/util/DownloadTest.java
```diff
@@ -32,15 +32,16 @@
 
     /**
      * Tests download of files - both bulk and concurrently.
+     *
      * @param uploadedChecksum checksums of the artifact uploaded to Artifactory, used to compare with the downloaded artifact.
```
yahavi, line 36:
```suggestion
     * @param uploadedChecksum - checksums of the artifact uploaded to Artifactory, used to compare with the downloaded artifact.
```

build-info-extractor/src/testFixtures/java/org/jfrog/build/IntegrationTestsBase.java
```diff
@@ -74,7 +76,7 @@ public void init() throws IOException {
             inputStream.close();
         }
 
-        url = readParam(props, "url");
+        url = Optional.ofNullable(readParam(props, "url")).orElse("");
```
yahavi, line 79: It looks like the new code is a dead one. `readParam` fails the test if the returned value is null.

build-info-extractor/src/testFixtures/java/org/jfrog/build/IntegrationTestsBase.java
```diff
@@ -306,4 +311,24 @@ public void setFiles(List<String> files) {
...
 
+    /***
+     * Returns a custom exception error to be thrown when repository creation/deletion fails.
+     * @param repo - repo name.
+     * @param creating - creation or deletion failed.
+     * @param statusCode - status code returned.
+     * @param reason - reason returned.
+     * @return - custom error message.
+     */
+    private String getRepositoryCustomErrorMessage(String repo, boolean creating, int statusCode, String reason) {
```
yahavi, line 322: I think using this syntax will be shorter:
```suggestion
    private String getRepositoryCustomErrorMessage(String repo, boolean creating, StatusLine statusLine) {
```

yahavi (changes requested):

build-info-extractor/src/testFixtures/java/org/jfrog/build/IntegrationTestsBase.java
RobiNino: True, but there is still a warning that url might be null. This was added to make sure it isn't and remove the warning.

yahavi (approved):

PR #437, opened 2021-01-03
Gradle Plugin - New "ALL_PUBLICATIONS" parameter

RobiNino:
-----

Following the discussion in #407, the _publications_ method now accepts a new parameter, "ALL_PUBLICATIONS", that will trigger the usage of all publications configured in the build script.
The parameter can currently be passed in a `build.gradle` file and used by the Artifactory Plugin directly, or be passed to the _publications_ API in Jenkins.
Other CI integrations will be supported in the future, after the _publications_ API will be implemented on their end.

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
@@ -11,6 +11,7 @@ import org.gradle.api.artifacts.repositories.IvyArtifactRepository
 import org.gradle.api.component.SoftwareComponent
 import org.gradle.api.invocation.Gradle
 import org.gradle.api.publish.Publication
+import org.gradle.api.publish.PublicationContainer
 import org.gradle.api.publish.PublishingExtension
 import org.gradle.api.publish.ivy.IvyPublication
 import org.gradle.api.publish.ivy.plugins.IvyPublishPlugin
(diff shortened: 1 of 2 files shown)


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
```diff
@@ -131,8 +132,16 @@ public class ProjectsEvaluatedBuildListener extends BuildAdapter implements Proj
      * @param publicationsNames - Publications names separated by commas
      */
     private static void addPublications(ArtifactoryTask artifactoryTask, PublishingExtension publishingExtension, String publicationsNames) {
+        PublicationContainer container = publishingExtension.getPublications();
         for (publicationName in publicationsNames.split(",")) {
-            Publication publication = publishingExtension.getPublications().findByName(publicationName);
+            // If ALL_PUBLICATIONS parameter was provided, add all publications and return.
+            if (publicationName == TaskHelperPublications.ALL_PUBLICATIONS) {
```
eyalbe4, line 138: I'm not sure if the following may introduce an issue or not, but even if doesn't, it may cause issues in the future, following changes to the code, so I syggest changing it to be on the safe side.
This code may cause the same publication to be added multiple times to the Artifactory Plugin's publications collection.
I suggest storing all the publications collected in the case of ALL_PUBLICATIONS in a different collction, then the collection used for storing the publications which names were specified specifically. This should allow saving either the ALL_PUBLICATIONS publications, or the specific ones.

eyalbe4 (changes requested):
I suggest adding to this PR description a usage example fo the new ALL_PUBLICATIONS option. This will make it easier for us to document this addition when this gets released.

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
RobiNino: I improved the code there to avoid redundant additions of publications when ALL_PUBLICATIONS is passed.
The publications are kept in a set so will be unique either way.
I think adding another collection will complicate the code a bit, since currently the CI and plugin share the same collections which is convenient.

PR #439, opened 2021-01-07
BI-557 - Populate requestedBy field in Maven extractor

yahavi:
-----

Listen to DependencyResolutionResult events and populate the `requiredBy` lists in the `Dependency` of the build info object.


eyalbe4 (changes requested):
See my above comments.

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/DependencyResolutionSpy.java
```diff
@@ -0,0 +1,104 @@
...
+import org.codehaus.plexus.component.annotations.Requirement;
+import org.eclipse.aether.artifact.Artifact;
+import org.eclipse.aether.graph.DependencyNode;
+import org.jfrog.build.extractor.BuildInfoExtractorUtils;
+
+import java.util.*;
+
+/**
+ * Listen to DependencyResolutionResult events and populate parents map for each dependency of the current module.
+ * Each parent is an array of strings contains the path from parent GAV to module GAV.
```
eyalbe4, line 17: Each parent is an array of strings contains the path from parent GAV to module GAV.
-->
Each parent is an array of strings, containing the path from parent GAV to module GAV.

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/DependencyResolutionSpy.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public void onEvent(Object event) {
+        if (event instanceof DependencyResolutionResult) {
+            DependencyResolutionResult result = (DependencyResolutionResult) event;
+            buildInfoRecorder.setDependencyParentsMaps(createDependencyParentsMap(result.getDependencyGraph()));
+        }
+    }
+
+    /**
+     * Create a map of dependency to parents.
```
eyalbe4, line 37: Let's ellaborate more to make it as easy as possble for the random reader to understand how the returned map is structured.
Let's also add an example, and use ti to exp,ain the structure.

eyalbe4 (changes requested):
Also see comments from the initial review.

PR #448, opened 2021-01-27
BI-560 - Add branch to Vcs

yahavi:
-----


RobiNino (approved):
You can also add `branch` to `sendBuildInfoTest()` if you'd like.

yahavi:
> You can also add `branch` to `sendBuildInfoTest()` if you'd like.

I can't because the sendBuildInfoTest sends the build info to Artifactory and compare the results. 
Since Artifactory does not yet support the branch field, we will be able to add more tests only after the next Artifactory release.

PR #452, opened 2021-02-03
BI-562 - Command executor in Linux does not accept multiple commands

yahavi:
-----

Resolves https://github.com/jfrog/jenkins-artifactory-plugin/issues/415

Currently the command executor for Linux, run each command separately:
```java
Runtime.getRuntime().exec(args.toArray(new String[0]), env, execDir)
```
However, it doesn't work for multiple commands, like commands with `&&`.

To support multiple commands, the command executor can get for both Linux and Mac 3 arguments:
1. `/bin/bash`
2. `-c`
3. The commands to run in 1 string

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
@@ -157,11 +157,10 @@ public class CommandExecutor implements Serializable {
         }
         if (SystemUtils.IS_OS_WINDOWS) {
             args.addAll(0, Arrays.asList("cmd", "/c"));
-        } else if (SystemUtils.IS_OS_MAC) {
-            // In MacOS, the arguments for '/bin/sh -c' must be a single string. For example "/bin/sh","-c", "npm i".
+        } else {
             String strArgs = join(" ", args);
             args = new ArrayList<String>() {{
-                add("/bin/sh");
+                add("/bin/bash");
                 add("-c");
                 add(strArgs);
             }};


eyalbe4 (approved):

build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -157,11 +157,10 @@ private static Process runProcess(File execDir, List<String> args, List<String>
         }
         if (SystemUtils.IS_OS_WINDOWS) {
             args.addAll(0, Arrays.asList("cmd", "/c"));
-        } else if (SystemUtils.IS_OS_MAC) {
-            // In MacOS, the arguments for '/bin/sh -c' must be a single string. For example "/bin/sh","-c", "npm i".
+        } else {
             String strArgs = join(" ", args);
             args = new ArrayList<String>() {{
-                add("/bin/sh");
+                add("/bin/bash");
```
saper, line 163: This is not portable really... what about /bin/sh here?

saper, line 163: Why change it to bash? Many systems (like FreeBSD) do not have `bash` here. I'd stay with `/bin/sh` here

yahavi: @saper
Thanks for your important feedback.

In some systems, like Ubuntu, the default shell is Dash. Dash does not support the source command:
image
We changed to bash to support the `source` command. The source command has to be supported to activate Python Virtualenvs. 

It is clear now that to support all scenarios, we have to think about a better solution before the release.

saper: We don't really know what can be invoked here at this point, do we?

I think the way the virtualenv support has been implemented is kind of fishy. Injecting a shell compound command like this can break in many interesting ways. Maybe the activate command should be run directly and only the state of the environment should be recorded by build info afterwards. This could allow us to activate the environment once for other commands, too. But that's a deeper story probably.

But back to the point: Can we try using shell "." (dot command) instead of "source"? It works with all shells pretending to be compatible with the Bourne shell.

yahavi: @saper 
Thanks again for your feedback.
I created https://github.com/jfrog/build-info/pull/453 to change `/bin/bash` to `/bin/sh`. 

I'm not sure how can we activate the virtualenv before running the pip install.
This command downloads dependencies from Artifactory, therefore it should get the artifactory connection details.
This command also collects the build info dependencies by parsing the logs during the pip install command, therefore it requires wrapping the pip install command.
To support running it in Docker images, all of the above run in a separated Java process.

```groovy
// Configure pip build
def rtPip = Artifactory.newPipBuild()
rtPip.resolver repo: 'pypi-virtual', server: server
def virtual_env_activation = ". /Users/myUser/venv-example/bin/activate"

// Run the pip install in a new Java process
def buildInfo = rtPip.install args: "-r python-example/requirements.txt", envActivation: virtual_env_activation
```

Please let us know if you have any ideas.

saper: @yahavi I must admit I have trouble following buldinfo's and the plugins code... so many layers. But I noticed there is some part of the BuildInfo that collects and manages the environment variables. Here is what could be done: provide an somewhere (rtPipResolver? separate command?) to just give a link to the directory (without a dot and activate).

The code should run the shell with a dot added AND record new variables added to the environment. It could either be permanent or maybe we should give the user a choice where to activate the environment (as they might want multiple ones)

```groovy
def rtPip = Artifactory.newPipBuild()
rtPip.resolver repo: 'pypi-virtual', server: server, virtualEnvironment: "/Users/myUser/venv-example"

// Run the pip install in a new Java process
def buildInfo = rtPip.install args: "-r python-example/requirements.txt", virtualEnvironment: "/Users/myUser/venv-example"
```

On the implementation side every virtual environment could be a Map of environment variables that just replaces or augments the `env` parameter of `PipInstall`.

PR #457, opened 2021-03-10
BI-565 collect issues should ignore errors when the revision of the previous build info doesn't exist

Or-Geva:
-----

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/issuesCollection/IssuesCollector.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/issuesCollection/IssuesCollector.java
@@ -32,7 +32,16 @@ public class IssuesCollector implements Serializable {
     private static final String LATEST = "LATEST";
     private static final String GIT_LOG_LIMIT = "100";
 
+    public static Pattern REVISION_NOT_EXIST;
+
     public IssuesCollector() {
+    }
+
+    public static Pattern getRevisionNotExistPattern() {
+        if (REVISION_NOT_EXIST == null) {
+            REVISION_NOT_EXIST = Pattern.compile("fatal: Invalid revision range [a-fA-F0-9]+\\.\\.");
+        }
+        return REVISION_NOT_EXIST;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/issuesCollection/IssuesCollector.java
```diff
@@ -141,6 +141,11 @@ private String getGitLog(File execDir, Log logger, String previousVcsRevision) t
         CommandExecutor commandExecutor = new CommandExecutor("git", null);
         CommandResults res = commandExecutor.exeCommand(execDir, args, null, logger);
         if (!res.isOk()) {
+            Pattern pattern = Pattern.compile("fatal: Invalid revision range [a-fA-F0-9]+\\.\\.");
```
eyalbe4, line 144: Since the ```Pattern.compile``` action is expensive, we'd like to avoid executing it everytime. Let's compile the pattern once, the first time it is needed and save it. When implementing this change, let's also avoid compiling the pattern, if it is not used.

build-info-extractor/src/test/java/org/jfrog/build/extractor/issuesCollection/IssuesCollectorTest.java
```diff
@@ -92,6 +92,13 @@ public void testCollectIssuesWithRevision() throws IOException, InterruptedExcep
         runCollectIssues("git_issues2_.git_suffix", vcsList, 2);
     }
 
+    @Test
+    // Test collection with a made up revision - the command should not throw an error, and 0 issues should be returned.
+    public void testCollectIssuesWithBadRevision() throws IOException, InterruptedException {
```
eyalbe4, line 97: testCollectIssuesWithBadRevision --> testCollectIssuesWithNonExistingRevision

build-info-extractor/src/test/java/org/jfrog/build/extractor/issuesCollection/IssuesCollectorTest.java
```diff
@@ -92,6 +92,13 @@ public void testCollectIssuesWithRevision() throws IOException, InterruptedExcep
         runCollectIssues("git_issues2_.git_suffix", vcsList, 2);
     }
 
+    @Test
+    // Test collection with a made up revision - the command should not throw an error, and 0 issues should be returned.
```
eyalbe4, line 96: Let's change the comment style to be 
```
/**

 */
```
Let's also add to the description the reason why we test this scenario.

eyalbe4 (changes requested):

PR #491, opened 2021-05-12
Bugfix - env not collect in npm, Go, Pip, and NuGet builds

yahavi:
-----

When building npm, Go, Pip, and NuGet projects, we should allow collecting the environment during the extractor process.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerExtractor.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerExtractor.java
@@ -35,6 +35,7 @@ public abstract class PackageManagerExtractor implements Serializable {
         if (build == null) {
             return;
         }
+        PackageManagerUtils.collectEnvIfNeeded(clientConfiguration, build);
         saveBuildInfoToFile(clientConfiguration, build);
     }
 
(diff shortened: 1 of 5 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerExtractor.java
```diff
@@ -56,6 +60,28 @@ static void saveBuildInfoToFile(ArtifactoryClientConfiguration clientConfigurati
         }
     }
 
+    /**
+     * Collect environment variables according to the env include-exclude patterns
+     *
+     * @param clientConfiguration - Artifactory client configuration
+     * @param build               - The target build
+     */
+    private static void collectEnv(ArtifactoryClientConfiguration clientConfiguration, Build build) {
```
eyalbe4, line 69: I think it is worth moving this method into a util, and have the maven and gradle flow use it as well, instead of the separate logic used by them. I understand that this will require some refactoring for maven and gradle, but I think the code will benefit from this. We can then test the collect env vars functionality from the Jenkins code base, to make sure we're good.

eyalbe4 (changes requested):

eyalbe4 (approved):

PR #501, opened 2021-05-25
Fix deserialization error in GetRepositoriesKeys

yahavi:
-----

The response of `GET api/repositories`is a list of repositories. Therefore the response value of GetRepositoriesKey should be a list.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/artifactory/services/GetRepositoriesKeys.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/artifactory/services/GetRepositoriesKeys.java
@@ -1,5 +1,6 @@
 package org.jfrog.build.extractor.clientConfiguration.client.artifactory.services;
 
+import com.fasterxml.jackson.core.type.TypeReference;
 import org.apache.http.HttpEntity;
 import org.apache.http.client.methods.HttpGet;
 import org.apache.http.client.methods.HttpRequestBase;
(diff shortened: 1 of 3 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/artifactory/services/GetRepositoriesKeys.java
```diff
@@ -37,8 +38,9 @@ public HttpRequestBase createRequest() {
 
     @Override
     protected void setResponse(InputStream stream) throws IOException {
-        GetRepositoriesKeyResponse localRepositories = getMapper(true).readValue(stream, GetRepositoriesKeyResponse.class);
-        result = localRepositories.getRepositoriesKey();
+        List<GetRepositoriesKeyResponse> keys = getMapper(true).readValue(stream, new TypeReference<List<GetRepositoriesKeyResponse>>() {
+        });
```
Or-Geva, line 42: Let's do the same as we do in RetrievePatternArtifacts
```suggestion
        List<GetRepositoriesKeyResponse> keys = getMapper(true).readValue(stream, TypeFactory.defaultInstance().constructCollectionLikeType(List.class, GetRepositoriesKeyResponse.class));
```

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/response/GetRepositoriesKeyResponse.java
```diff
@@ -1,32 +1,13 @@
 package org.jfrog.build.extractor.clientConfiguration.client.response;
 
-import java.util.List;
-import java.util.stream.Collectors;
-
 public class GetRepositoriesKeyResponse {
```
Or-Geva, line 3: `GetRepositoriesKeyResponse` -> `GetRepositoriesResponse`
Right now we are only looking for keys but, this can be used by other types of 'get repositories' services in the future.

eyalbe4 (approved):

PR #502, opened 2021-05-26
Fix: Ignore missing fields while deserialize HTTP response

gailazar300:
-----


yahavi (changes requested):
I think it would be better to remove the `ignoreMissingFields` at all and always ignore missing fields.

yahavi (approved):

PR #507, opened 2021-05-28
Bugfix - Error when trying to download an empty (zero bytes size) file

Or-Geva:
-----

All JFrog services ignore response with zero content length, this isn't fit to upload service, as it stops the download flow before reaching file creation. Therefore it was removed.
Fixes  https://github.com/jfrog/jenkins-artifactory-plugin/issues/490

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/JFrogService.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/JFrogService.java
@@ -149,8 +149,7 @@ public abstract class JFrogService<TResult> {
             return;
         }
         try (InputStream stream = entity.getContent()) {
-            long contentLength = entity.getContentLength();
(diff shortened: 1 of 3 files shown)


Or-Geva:
> ```
> I think we there is no point in creating a new input stream if the response type is `EMPTY`.
> ```
You're right.

> ```
> Moreover, shouldn't we call handleEmptyEntity() in that case?
> ```
No, `JFrogServiceResponseType.EMPTY` means that the service expects nothing to be returned (void) and supposed return successfully, whereas `handleEmptyEntity` handles a service that expects to find one but got no entity.
@yahavi, I propose the following:
```java
private void processResponse(HttpEntity entity) throws IOException {
    if (responseType == JFrogServiceResponseType.EMPTY) {
        return;
    }
    if (entity == null ) {
        handleEmptyEntity();
    }
    try (InputStream stream = entity.getContent()) {
        setResponse(stream);
    }
}
```

yahavi:
@Or-Geva OK, but please add Javadoc above `handleEmptyEntity` explaining when it should and shouldn't be called.

Issue #149, opened 2018-02-05
Honour Java system proxy settings


mithomas:
Currently the gradle plugin does no support the default Java proxy settings http(s).proxyHost/http(s).proxyPort/http(s).nonProxyHosts.

Even though it is possible to pass through the settings like so:
```
artifactory {
  clientConfig.proxy.host = System.properties['http.proxyHost']
  clientConfig.proxy.port = System.properties['http.proxyPort'] as Integer
}
```
this falls short for non-proxy-hosts or for different http/https proxies.

PR #510, opened 2021-06-01
Honor system proxy settings

drewhamilton:
-----

This partially addresses #149. If no `proxyHost` is specified by the Gradle plugin consumer, the Gradle plugin tries using the JVM "http(s).proxyHost" property. Similarly, if no `proxyPort` is specified by the Gradle plugin consumer, it tries using the JVM "http(s).proxyPort" property.

Whether the "http" or "https" property is used depends on the `contextUrl`: "http" if the `contextUrl` starts with "http://", "https" otherwise.

This does not address the part of #149 that mentions non-proxy-hosts. **This may be a blocker for this PR**, since now this plugin could try using a proxy when `contextUrl` is actually part of non-proxy-hosts. Please let me know if you think non-proxy-hosts handling must be part of this PR.

I didn't write tests yet since I'd like confirmation about whether the above behavior choices are correct. Once these are confirmed I'm happy to add tests.

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/DeployTask.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/DeployTask.java
@@ -178,7 +178,7 @@ public class DeployTask extends DefaultTask {
                             IncludeExcludePatterns patterns = new IncludeExcludePatterns(
                                     publisher.getIncludePatterns(),
                                     publisher.getExcludePatterns());
-                            configureProxy(accRoot, artifactoryManager);
+                            configureProxy(contextUrl, accRoot, artifactoryManager);
                             configConnectionTimeout(accRoot, artifactoryManager);
                             configRetriesParams(accRoot, artifactoryManager);
                             deployArtifacts(artifactoryTask.deployDetails, artifactoryManager, patterns, logPrefix, publisher.getMinChecksumDeploySizeKb());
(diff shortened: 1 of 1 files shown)


eyalbe4:
Thanks for this contribution @drewhamilton!
Looks good to me.
Let's just add a unit test for the ```configureProxy``` method.
Also, please go ahead and sign the CLA as requested in the above comment.

drewhamilton:

eyalbe4:
I'll have a look at the code soon @drewhamilton.
Yes - let's make the method public for the test.

eyalbe4:
@drewhamilton - 
It looks like this PR was mistakenly skipped. 
Please accept my apologies for this...
Would you like us to review the PR again? Are you still interested in promoting this functionality?

drewhamilton:
@eyalbe4 I don't personally need this anymore, though as far as I know it's still a valid approach.

@paristote Would this be helpful for your CI/Gradle setup?

paristote:
Thanks for coming back to this PR, @eyalbe4 , @drewhamilton. Yes this is still useful for us, without it we need extra configuration to set the proxy settings for this particular plugin.

build-info-extractor-gradle/src/test/java/org/jfrog/gradle/plugin/artifactory/task/DeployTaskProxyTest.java
```diff
@@ -0,0 +1,242 @@
...
+        assertEquals(result.host, "proxy-host");
+        assertEquals(result.port, 9999);
+    }
+
+    @Test
+    public void givenSystemProxyHostAndPort_whenContextUrlIsHttp_usesSystemHttpProxyHostAndPort() {
+        System.setProperty(HTTPS_PROXY_HOST_SETTING, "https-proxy-host");
+        System.setProperty(HTTPS_PROXY_PORT_SETTING, String.valueOf(1111));
+        System.setProperty(HTTP_PROXY_HOST_SETTING, "http-proxy-host");
+        System.setProperty(HTTP_PROXY_PORT_SETTING, String.valueOf(2222));
```
eyalbe4, line 134: The above line are duplicated many times in the code. I suggest we move them into a method and reuse it.

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/task/DeployTask.java
```diff
@@ -201,19 +202,88 @@ private void deployArtifacts(ArtifactoryClientConfiguration accRoot, Map<String,
...
+    private static boolean matchesProxyHostWildcardPattern(@Nonnull String string, @Nonnull String pattern) {
+        String[] patternParts = Arrays.stream(StringUtils.split(pattern, "*"))
+                // Drop leading '.', so that string with "example.com" will match pattern part ".example.com":
+                .map(it -> it.startsWith(".") ? it.substring(1) : it)
+                .toArray(String[]::new);
+
+        if (!pattern.startsWith("*") && !string.startsWith(patternParts[0])) {
+            // Shortcut loop if string doesn't start with an exact match:
+            return false;
+        } else {
```
eyalbe4, line 277: This "else" clause is redundant, because of the above "return".

eyalbe4 (changes requested):

PR #531, opened 2021-07-09
Change default Gradle publications

yahavi:
-----

Resolves https://github.com/jfrog/jenkins-artifactory-plugin/issues/500
Resolves https://github.com/jfrog/jenkins-artifactory-plugin/issues/508

When the [war](https://docs.gradle.org/current/userguide/war_plugin.html) plugin applied, JARs doesn't produce by default, but only WARs:
> The War plugin extends the Java plugin to add support for assembling web application WAR files. It disables the default JAR archive generation of the Java plugin and adds a default WAR archive task.

This PR Introduces the following:
1. New default `mavenWeb` publication which may be added only if the war plugin applied and enabled:
```groovy
mavenWeb(MavenPublication)
    from components.web
}
```
2. The default `mavenJava` and `ivyJava` publications wouldn't be added if the jar task is disabled.


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
```diff
@@ -170,15 +186,27 @@ public class ProjectsEvaluatedBuildListener extends BuildAdapter implements Proj
...
-            addIvyJavaPublication(publishingExtension, artifactoryTask.project)
+            // Add mavenWeb publication if war task exists and enabled
+            Task warTask = artifactoryTask.project.tasks.findByName("war");
+            if (warTask != null && warTask.enabled) {
+                addMavenWebPublication(publishingExtension, artifactoryTask.project)
+            }
+
+            // Add mavenJava and ivyJava publications if jar task doesn't exist, or exists and enabled
+            Task jarTask = artifactoryTask.project.tasks.findByName("jar");
+            if (jarTask == null || (jarTask != null && jarTask.enabled)) {
```
eyalbe4, line 204: This if statement can be simplified to:
```
if (jarTask == null || jarTask.enabled) {
```

eyalbe4 (changes requested):

PR #534, opened 2021-07-13
Switch to case insensitive comparison of HTTP headers

### LINKED ISSUE #533: DependenciesDownloadHelper#downloadArtifactMetaData() handles header case incorrectly
**Describe the bug**
Found this problem while upgrading Jenkins Artifactory plugin from 3.11.0 to 3.11.1. It in turn switches from version 2.26.0 to 2.26.2 of this library.
In 67cefb2 the method downloadArtifactMetaData in DependenciesDownloadHelper is refactored for some reason. Unfortunately the refactoring leads to incorrect handling of HTTP headers. Previously the header name was treated case-insensitively, but after the change it requires a certain case of the header name. If the header name doesn't mach, then, the file isn't downloaded due to the code in downloadArtifact() (which assumes that an artifact is a folder if no md5 or Sha1 is found).

Normally this isn't a problem since Artifactory produces the headers in expected case (X-Checksum-Md5 and X-Checksum-Sha1). But, in our case we have an HAProxy in between which transforms all headers to lower case (due to HTTP2 compliance). And thus the Jenkins Artifactory plugin fails to download the artifacts since the header isn't found by this library.

To be honest I don't see the point in the refactoring done to downloadArtifactMetaData(). The previous implementation relied on HttpResponse#getFirstHeader() to resolve header name to header value, which works perfectly fine. HttpResponse is a well tested and trusted implementation of header handling. Implementing your own header handling is, in my opinion, just unwise!

**To Reproduce**
Try to download artifacts using this library through a proxy which transforms the response headers to lowercase.

**Expected behavior**
That this library follows HTTP specification and ignores case of the HTTP response headers.

**Versions**
2.26.2 of this library
HAProxy 2.2

**Additional context**
It is possible to work around the issue by setting the following config parameters to HAProxy (i.e. it will ensure the headers have expected format:
```
default:
 option h1-case-adjust-bogus-client
global:
 h1-case-adjust accept-ranges Accept-Ranges
 h1-case-adjust content-disposition Content-Disposition
 h1-case-adjust content-length Content-Length
 h1-case-adjust content-type Content-Type
 h1-case-adjust last-modified Last-Modified
 h1-case-adjust x-artifactory-filename X-Artifactory-Filename
 h1-case-adjust x-artifactory-id X-Artifactory-Id
 h1-case-adjust x-artifactory-node-id X-Artifactory-Node-Id
 h1-case-adjust x-checksum-md5 X-Checksum-Md5
 h1-case-adjust x-checksum-sha1 X-Checksum-Sha1
 h1-case-adjust x-checksum-sha256 X-Checksum-Sha256
 h1-case-adjust x-jfrog-version X-JFrog-Version
```
This is not a permanent fix though. This library should really follow the HTTP standard and ignore header case.

jonsten:
Fixes #533 

A bug was introduced in 67cefb2 where http headers no longer was compared case insensitively. This causes problems for Artifactory instances that run behind proxys that normalizes the http headers.

This change fixes the issue by mimicking the implementation in org.apache.http.message.HeaderGroup by using String#equalsIgnoreCase(). Which is the old and correct behaviour.

I haven't added any tests nor tried to build the code base since Gradle is unable to find some dependencies :(

-----

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/DependenciesDownloaderHelper.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/DependenciesDownloaderHelper.java
@@ -349,19 +349,14 @@ public class DependenciesDownloaderHelper {
         try {
             ArtifactMetaData artifactMetaData = new ArtifactMetaData();
             for (Header header : downloader.getArtifactoryManager().downloadHeaders(url)) {
-                switch (header.getName()) {
-                    case MD5_HEADER_NAME:
-                        artifactMetaData.setMd5(header.getValue());
-                        break;
-                    case SHA1_HEADER_NAME:
-                        artifactMetaData.setSha1(header.getValue());
-                        break;
-                    case HttpHeaders.CONTENT_LENGTH:
-                        artifactMetaData.setSize(NumberUtils.toLong(header.getValue()));
-                        break;
-                    case HttpHeaders.ACCEPT_RANGES:
-                        artifactMetaData.setAcceptRange("bytes".equals(header.getValue()));
-                        break;
+                if (MD5_HEADER_NAME.equalsIgnoreCase(header.getName())) {
+                    artifactMetaData.setMd5(header.getValue());
+                } else if (SHA1_HEADER_NAME.equalsIgnoreCase(header.getName())) {
+                    artifactMetaData.setSha1(header.getValue());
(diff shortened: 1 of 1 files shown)


jonsten:
Looks like we don't have any CLA signed. It will probably take a few weeks/months to get that setup through legal considering that it is vacation time right now...

yahavi:
@jonsten,
Thanks for your contribution!
Would you like that we will create a separate PR to fix this issue?

jonsten:
That is fine with me, thanks!

yahavi:
Thanks, @jonsten. I created #539 to fix this issue.

PR #545, opened 2021-08-08
Refactor build scan table

RobiNino:
-----
Screen Shot 2021-08-08 at 11 00 57

Separated to two tables, added CVE field


build-info-extractor/src/main/java/org/jfrog/build/extractor/buildScanTable/SecurityViolationsTable.java
```diff
@@ -0,0 +1,82 @@
...
+        if (infectedFile.getDisplayName() != null && infectedFile.getDisplayName().length() > longestDisplayName) {
+            longestDisplayName = infectedFile.getDisplayName().length();
+        }
+    }
+
+    void printTable() {
+        super.printTable(table);
+    }
+
+    static class SecurityTableElement extends TableElementBase {
```
eyalbe4, line 47: static class --> private static class

eyalbe4 (changes requested):

PR #546, opened 2021-08-09
Allow npm projects without name and versions

yahavi:
-----

`package.json` files are allowed to be without name and version nodes. Currently, the code throws NPE in such cases.
If `name` is missing, use the directory name as the project name.
If only `version` is missing, use the name without the version as the project name.
Otherwise, use <name>:<version> as the project name.

### DIFF
--- a/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmDependencyTree.java
+++ b/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmDependencyTree.java
@@ -8,6 +8,7 @@ import org.jfrog.build.extractor.npm.types.NpmScope;
 import org.jfrog.build.extractor.scan.DependencyTree;
 import org.jfrog.build.extractor.scan.Scope;
 
+import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmDependencyTree.java
```diff
@@ -21,20 +22,41 @@
...
     }
 
+    /**
+     * Get npm project name to populate the root node.
+     *
+     * @param npmList    - Results of 'npm ls' command
+     * @param workingDir - The package.json directory
+     * @return <name>:<version>, <name> or <directory-name>
+     */
+    static String getProjectName(JsonNode npmList, Path workingDir) {
```
eyalbe4, line 48: static String --> private static String

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmDependencyTree.java
```diff
@@ -21,20 +22,41 @@
...
+        JsonNode name = npmList.get("name");
+        JsonNode version = npmList.get("version");
+        if (name != null && version != null) {
+            return name.asText() + ":" + version.asText();
+        }
+        if (name != null) {
+            return name.asText();
+        }
+        return workingDir.getFileName().toString();
+    }
```
eyalbe4, line 58: This method code can be slightly simplified and changed to - 
```
        JsonNode name = npmList.get("name");
        JsonNode version = npmList.get("version");
        if (name != null) {
            if (version != null) {
                return name.asText() + ":" + version.asText();
            }
            return name.asText();
        }
        return workingDir.getFileName().toString();
```

eyalbe4 (changes requested):

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmDependencyTree.java
yahavi: But this function is used in the test

PR #552, opened 2021-09-13
Use the file-specs-java library

asafgabai:
-----


build.gradle
```diff
@@ -83,6 +83,7 @@ subprojects {
         implementation 'org.apache.httpcomponents:httpcore:4.4.5'
         implementation "commons-codec:commons-codec:1.13"
         implementation group: 'com.google.guava', name: 'guava', 'version': '30.0-jre'
+        implementation group: 'org.jfrog.filespecs', name: 'file-specs-java', version: '1.0.3'
```
asafgabai, line 86: Reminder to myself to update this after the library is released.

build-info-api/src/main/java/org/jfrog/build/api/util/Log.java
```diff
@@ -23,14 +23,14 @@
  *
  * @author Noam Y. Tenne
  */
-public interface Log extends Serializable {
-    public void debug(String message);
+public interface Log extends Serializable, org.jfrog.filespecs.utils.Log {
```
eyalbe4, line 26: Why do we not keep the methods of this interface public?

eyalbe4 (changes requested):

PR #559, opened 2021-10-07
Split build-info class

Or-Geva:
-----
Until now, Build-info-java has one class that includes all the build-info schema properties & extra logic. Examples of build-info logic are recording artifacts/dependencies local path or append modules/builds.

The one-class structure needs to be splitted into two classes. The first class would have only build-info properties that Artifactory will be using, and the second would have all the extra logic + properties which will be used by other consumers (Jenkins).


build-info-api/src/main/java/org/jfrog/build/api/builder/BuildBuilder.java
```diff
@@ -0,0 +1,403 @@
...
+        return this;
+    }
+
+    /**
+     * Sets the violation notifications of the build
+     *
+     * @param licenseControl Build violation  recipients.
+     * @return Builder instance
+     */
+    public BuildBuilder licenseControl(LicenseControl licenseControl) {
```
eyalbe4, line 356: This method, in addition to everything that is around license control should be removed. I suggest we delete the LicenseControl class, and then remove everything that uses it.

eyalbe4 (changes requested):
I'm not done reviewing everything, but I'm sharing what I have so far.
I'll continue reviewing the rest soon.

build-info-api/src/main/java/org/jfrog/build/api/builder/BuildInfoBuilder.java
```diff
@@ -81,40 +88,39 @@ public Build build() {
             throw new IllegalArgumentException("Build start time must be set");
         }
 
-        Build build = new Build();
+        BuildInfo buildInfo = new BuildInfo();
         if (StringUtils.isNotBlank(version)) {
```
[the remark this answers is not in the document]
Or-Geva: The If makes sure that the default version in `new BuildInfo();` won't get an override.

build-info-api/src/main/java/org/jfrog/build/api/builder/BuildInfoModuleBuilder.java
```diff
@@ -0,0 +1,217 @@
...
+import org.jfrog.build.api.ci.Artifact;
+import org.jfrog.build.api.ci.Dependency;
+import org.jfrog.build.api.ci.Module;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Properties;
+
+
+public class BuildInfoModuleBuilder {
```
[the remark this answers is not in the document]
Or-Geva: BuildModuleBuilder - build the old 'build' class (used by Artifactory)
BuildInfoModuleBuilder - build the new 'build' class (used by Jenkins, etc..)

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -1,37 +1,3 @@
...
- * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
- * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
- * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
- * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
- * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
- * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
- * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
- * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
- * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- */
```
eyalbe4, line 33: We need to keep this specific declaration.

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildDeploymentHelper.java
```diff
@@ -48,26 +36,26 @@
...
-    public void deploy( Build                          build,
-                        ArtifactoryClientConfiguration clientConf,
-                        Map<String, DeployDetails>     deployableArtifactBuilders,
-                        boolean                        wereThereTestFailures,
-                        File                           basedir ) {
+    public void deploy(BuildInfo buildInfo,
+                       ArtifactoryClientConfiguration clientConf,
+                       Map<String, DeployDetails> deployableArtifactBuilders,
+                       boolean wereThereTestFailures,
+                       File basedir) {
```
eyalbe4, line 43: Let's use this opportunity and align the above code lines according to our conventions.

eyalbe4 (changes requested):

PR #580, opened 2021-12-05
Update apache commons-lang (2) to current commons-lang3

ebariaux:
See issue https://github.com/jfrog/jenkins-artifactory-plugin/issues/556, PR https://github.com/jfrog/jenkins-artifactory-plugin/pull/567 was not sufficient to ensure commons-lang 2 is not included in the libs of the built plugin, it should also be removed from the dependant projects, including this one.


ebariaux:

yahavi (approved):
LGTM. Thanks!

ebariaux:
Today while performing further testing on our Jenkins, I had a failure because of this change.
The build-info-extractor-docker is relying on com.github.docker-java component, which is still using commons lang 2.
This caused an exception in the Jenkins plugin.
I've created a PR on docker-java to also migrate it to commons lang 3, see https://github.com/docker-java/docker-java/pull/1746
When referencing that updated version from this project and re-building the Jenkins Artifactory Plugin issue was solved.

PR #594, opened 2021-12-26
Add support for v2 compatible go projects with submodule as root

jfrog-qa:
-----

### DIFF
--- a/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
+++ b/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
@@ -41,11 +41,14 @@ import org.apache.commons.compress.archivers.zip.ZipArchiveOutputStream;
 import org.apache.commons.compress.archivers.zip.ZipFile;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.jfrog.build.api.util.Log;
 
 import java.io.Closeable;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
 import java.util.Enumeration;
 import java.util.Set;
 import java.util.zip.ZipEntry;
@@ -58,12 +61,13 @@ public class GoZipBallStreamer implements Closeable {
 
     protected ArchiveOutputStream archiveOutputStream;
     private final Log log;
-    private ZipFile zipFile;
-    private String projectName;
-    private String version;
-    private Set<String> excludedDirectories;
-    private String subModuleName;
-    private static final String MOD_FILE = "/go.mod";
+    private final ZipFile zipFile;
+    private final String projectName;
+    private final String version;
+    private final Set<String> excludedDirectories;
+    private String subModuleName = "";
+    private static final String MOD_FILE = "go.mod";
+    private static final String MOD_FILE_PATH = "/" + MOD_FILE;
     private static final String VENDOR = "vendor/";
 
     public GoZipBallStreamer(ZipFile zipFile, String projectName, String version, Log log) {
@@ -89,15 +93,22 @@ public class GoZipBallStreamer implements Closeable {
         writeEntries();
     }
 
+    void setSubModuleName(String subModuleName) {
+        this.subModuleName = subModuleName;
+    }
+
     /**
-     * Determine if the project is a compatible module from version 2+, a sub module or an incompatible module
+     * Determine if the project is a compatible module from version 2+, a submodule or an incompatible module
      */
     private void initiateProjectType() {
         boolean compatibleModuleFromV2 = GoVersionUtils.getMajorVersion(version, log) >= 2 &&
                 GoVersionUtils.isCompatibleGoModuleNaming(projectName, version, log);
         if (compatibleModuleFromV2) {
-            subModuleName = "v" + GoVersionUtils.getMajorProjectVersion(projectName, log);
-            log.debug(projectName + "@" + version + " is compatible Go module from major version " + subModuleName);
+            String majorVersion = "v" + GoVersionUtils.getMajorProjectVersion(projectName, log);
+            if (!hasRootModFileOfCompatibleModuleFromV2(majorVersion)) {
+                subModuleName = majorVersion;
+                log.debug(projectName + "@" + version + " is compatible Go module from major version " + subModuleName);
+            }
         } else {
             subModuleName = GoVersionUtils.getSubModule(projectName);
             if (shouldPackSubModule()) {
@@ -195,7 +206,7 @@ public class GoZipBallStreamer implements Closeable {
         while (entries.hasMoreElements()) {
             zipEntry = entries.nextElement();
             if (!zipEntry.isDirectory() && isSubModule(zipEntry.getName())) {
-                String subModulePath = zipEntry.getName().replace(MOD_FILE, "");
+                String subModulePath = zipEntry.getName().replace(MOD_FILE_PATH, "");
                 excludedDirectories.add(subModulePath);
             } else {
                 allDirectories.add(GoVersionUtils.getParent(zipEntry.getName()));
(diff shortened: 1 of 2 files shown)


build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -238,29 +244,59 @@ private boolean isSubFolderOfAnotherModule(String moduleRootDir, String director
...
-                return entryName.substring(entryName.indexOf('/') + 1).endsWith(subModuleName + MOD_FILE);
+                return !entryName.substring(entryName.indexOf('/') + 1).equals(MOD_FILE);
+            }
+        }
+        return false;
+    }
+
+    /**
+     * @param majorVersion - major version of a compatible v2 project
+     * @return true, if the go.mod file in the root specify a path of submodule of the given major version
```
eyalbe4, line 259: specify --> specifies

Since this description is not easy to understand, I suggest adding an example to the description.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -287,4 +323,25 @@ private static String stripFirstPathElement(String path) {
     public void close() throws IOException {
         zipFile.close();
     }
+
+    private String getFirstPathElement(String path) {
```
eyalbe4, line 327: Let's add a clear description for this method.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -287,4 +323,25 @@ private static String stripFirstPathElement(String path) {
...
+        if (path == null) {
+            return null;
+        } else {
+            String[] elements = getPathElements(path);
+            return elements.length > 0 ? elements[0] : "";
+        }
+    }
+
+    private String[] getPathElements(String path) {
```
eyalbe4, line 336: Let's add a clear description for this method.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -238,29 +244,59 @@ private boolean isSubFolderOfAnotherModule(String moduleRootDir, String director
      * @return True if the entry is a go.mod file which doesn't belong to the project we are packing
      */
     private boolean isSubModule(String entryName) {
```
eyalbe4, line 246: Let's add a unit test for this method.

eyalbe4 (changes requested):

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
yahavi, line 327: I think all of this function can be replaced by:
```java
StringUtils.substringBefore(path, "/");
```
And a question - Can windows paths with `\` as a separator be here?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -238,29 +244,59 @@ private boolean isSubFolderOfAnotherModule(String moduleRootDir, String director
...
+        return false;
+    }
+
+    /**
+     * @param majorVersion - major version of a compatible v2 project
+     * @return true, if the go.mod file in the root specify a path of submodule of the given major version
+     */
+    private boolean hasRootModFileOfCompatibleModuleFromV2(String majorVersion) {
+        Enumeration<? extends ZipArchiveEntry> entries = zipFile.getEntries();
+        if (entries.hasMoreElements()) {
```
yahavi, line 263: Shouldn't be `while`?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -238,29 +244,59 @@ private boolean isSubFolderOfAnotherModule(String moduleRootDir, String director
...
+            ZipArchiveEntry zipEntry = entries.nextElement();
+            ZipArchiveEntry modEntry = zipFile.getEntry(getFirstPathElement(zipEntry.getName()) + MOD_FILE_PATH);
+            if (modEntry == null) {
+                modEntry = zipFile.getEntry(MOD_FILE);
+            }
+            if (modEntry != null) {
+                try (InputStream inputStream = zipFile.getInputStream(modEntry)) {
+                    String modFileContent = IOUtils.toString(inputStream, Charset.defaultCharset());
+                    return modFileContent.substring(0, modFileContent.indexOf("\n")).endsWith("/" + majorVersion);
```
yahavi, line 272: I'm not sure it will work on Windows, whereby in some cases the new line character is `\r\n`.
Let's trim the string before calling `endsWith`.
Also - if we replace `substring` with `StringUtils.substringBefore`, the code wouldn't crash on empty go.mod files.

yahavi (changes requested):

yahavi:
Please also rename the title of the PR to something more informative than `Bugfix/rtdev 3462`

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -238,29 +244,59 @@ private boolean isSubFolderOfAnotherModule(String moduleRootDir, String director
...
+                } catch (IOException e) {
+                    log.debug("Failed to read go.mod file of the root project");
```
yahavi, line 274: Let's add more information about the failed entry, so we can fix bugs. The information should include at least the full path in the archive to the entry.
Also, please consider changing the `debug` log level to `info` or `warning`, because this is an unexpected error.

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
jfrog-qa: Nope - no need to iterate on the zip file - we just want to check the content of the go.mod in the root (if exists)

yahavi: OK. In that case, could you please write this as a comment above the `if`?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
jfrog-qa: The full path is the go.mod for the root project (like i mentioned).
I can put it as info, and also add log.debug with the exception error. is it ok?

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
jfrog-qa: Nope - if path is "/example/go.mod" -> return value is "/"

yahavi: Got it, and still, those methods looks a bit complicated for me. 
Please consider the following:
```java
private String getFirstPathElement(String path) {
    if (path == null) {
        return "";
    }
    path = StringUtils.removeStart(path, "/");
    return StringUtils.substringBefore(path, "/");
}
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
yahavi: I'm not sure that the `debug` log level is enough. If this exception is unexpected we should print it to the user, so he can report it to us. 
You can add it as an error log message:
```java
log.error("Failed to read go.mod file of the root project", e);
```
Another option is to print the root cause message:
```java
log.warn("Failed to read go.mod file of the root project: " + ExceptionUtils.getRootCauseMessage(e));
```

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
jfrog-qa: ok NP

PR #595, opened 2021-12-28
Fix maven late deploy

Or-Geva:
-----
`isPublishArtifacts` disable build-info from generating `deployable.artifacts.json`.
This file is used by maven late artifacts deploy to know which artifact to deploy to where.

This PR replace isPublishArtifacts by adding a new property instead.

Related PR: https://github.com/jfrog/build-info-go/pull/26


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/ArtifactoryClientConfiguration.java
```diff
@@ -454,6 +455,10 @@ public Boolean isPublishArtifacts() {
             return getBooleanValue(PUBLISH_ARTIFACTS, true);
         }
 
+        public Boolean generateDeployArtifactsMap() {
```
talarian1, line 458: Maybe better to rename generateDeployArtifactsMap to shouldAddDeployableArtifacts ?

PR #612, opened 2022-02-02
Verify artifact type limit and add project support for maven

sverdlov93:
-----
Verify that the artifact type is less than 64 characters.
If it's longer, don't use classifier or extension as type.

Fixing https://github.com/jfrog/artifactory-maven-plugin/issues/29

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
@@ -37,6 +37,8 @@ import java.util.function.Predicate;
  */
 public abstract class BuildInfoExtractorUtils {
 
+    private static final int ARTIFACT_TYPE_LENGTH_LIMIT = 64;
+
     public static final Predicate<Object> BUILD_INFO_PREDICATE =
             new PrefixPredicate(BuildInfoProperties.BUILD_INFO_PREFIX);
 
@@ -277,6 +279,10 @@ public abstract class BuildInfoExtractorUtils {
             if (StringUtils.isNotBlank(extension) && !result.endsWith(extension)) {
                 result = result + "-" + extension;
             }
+        }
+         if (result.length() > ARTIFACT_TYPE_LENGTH_LIMIT) {
+             //  Artifactory limit for type length is 64
+            return type;
         }
         return result;
     }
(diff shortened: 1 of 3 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -277,6 +279,10 @@ public static String getTypeString(String type, String classifier, String extens
             if (StringUtils.isNotBlank(extension) && !result.endsWith(extension)) {
                 result = result + "-" + extension;
             }
+        }
+         if (result.length() > ARTIFACT_TYPE_LENGTH_LIMIT) {
+             //  Artifactory limit for type length is 64
```
eyalbe4, line 284: Redundant white-space after //

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -277,6 +279,10 @@ public static String getTypeString(String type, String classifier, String extens
             if (StringUtils.isNotBlank(extension) && !result.endsWith(extension)) {
                 result = result + "-" + extension;
             }
+        }
+         if (result.length() > ARTIFACT_TYPE_LENGTH_LIMIT) {
```
eyalbe4, line 283: This line should go one space back (it's not aligned with the rest of the code.

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -277,6 +279,10 @@ public static String getTypeString(String type, String classifier, String extens
...
+             //  Artifactory limit for type length is 64
+            return type;
         }
         return result;
```
eyalbe4, line 287: The above code - 
```
         if (result.length() > ARTIFACT_TYPE_LENGTH_LIMIT) {
             //  Artifactory limit for type length is 64
            return type;
        }
        return result;
```
to be - 
```
// Artifactory limit for type length is 64
return result.length() > ARTIFACT_TYPE_LENGTH_LIMIT) ? type : result;
```

eyalbe4 (changes requested):

PR #613, opened 2022-02-03
Support build promotion with projects

yahavi:
-----

Resolves https://github.com/jfrog/jenkins-artifactory-plugin/issues/627

### DIFF
--- a/build-info-extractor/src/test/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryManagerTest.java
+++ b/build-info-extractor/src/test/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryManagerTest.java
@@ -2,16 +2,11 @@ package org.jfrog.build.extractor.clientConfiguration.client;
 
 import org.apache.commons.io.FileUtils;
 import org.jfrog.build.IntegrationTestsBase;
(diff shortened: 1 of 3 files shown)


build-info-extractor/src/test/java/org/jfrog/build/extractor/clientConfiguration/client/ArtifactoryManagerTest.java
```diff
@@ -51,18 +46,49 @@ protected void cleanup() throws IOException {
...
     @Test
-    public void sendBuildInfoAndBuildRetentioTest() throws IOException {
-        doSendBuildInfoTest(null);
-        sendBuildRetention("");
+    public void sendBuildInfoAndBuildRetentionTest() throws IOException {
+        try {
+            publishAndCheckBuildInfo(null);
+            sendBuildRetention("");
+        } finally {
+            cleanTestBuilds(BUILD_NAME, BUILD_NUMBER, null);
```
eyalbe4, line 54: We send here ```null``` for the projecr argument, while we also send an empty string for this argument on other calls for this method. If null works, we better use it instead of an empty string.

eyalbe4 (changes requested):

PR #617, opened 2022-02-15
Fix Docker module name in build-info

yahavi:
-----

**Background**
Currently, there are at least 2 ways to configure the Docker registry URL: 
1. Using a reverse proxy or using the full path to the repository: acme-docker-local.jfrog.io. 
2. Using URL to the Docker repository: acme.jfrog.io/docker-local

**Expected**
In the generated build-info, the module ID should be `<image-name>/<image-tag>`.

**Actual**
When configuring URL to Docker repository as a Docker registry, the module ID, in this case: `docker-local/<image-name>/<image-tag>`.


eyalbe4 (approved):

gangefors:
This pull request broke the module id name for any image that contains more than one slash.

See https://github.com/jfrog/build-info/issues/665

PR #618, opened 2022-02-16
Improve support for Maven projects and fix build-info URL

yahavi:
-----

Fix https://github.com/jfrog/jenkins-artifactory-plugin/issues/576
Fix https://github.com/jfrog/jenkins-artifactory-plugin/issues/611

To support using build projects in Maven UI jobs in Jenkins Artifactory plugin, this PR does the following:
* Add the project in BuildInfoModelPropertyResolver.
* Fix the created build-info URL when using projects in all cases.
  *  If Artifactory's URL and the project key were provided, the `createBuildInfoUrl` will try to guess the platform URL from Artifactory URL.
  * The URL contained `?project=<project-key>` query param instead of `?buildRepo=<project-key>-build-info` param.

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoModelPropertyResolver.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoModelPropertyResolver.java
@@ -7,33 +7,19 @@ import org.apache.maven.execution.ExecutionEvent;
 import org.codehaus.plexus.component.annotations.Component;
 import org.codehaus.plexus.component.annotations.Requirement;
 import org.codehaus.plexus.logging.Logger;
-import org.jfrog.build.extractor.builder.BuildInfoMavenBuilder;
 import org.jfrog.build.api.builder.PromotionStatusBuilder;
-import org.jfrog.build.extractor.ci.Agent;
-import org.jfrog.build.extractor.ci.BuildAgent;
-import org.jfrog.build.extractor.ci.BuildInfo;
-import org.jfrog.build.extractor.ci.Issue;
-import org.jfrog.build.extractor.ci.IssueTracker;
-import org.jfrog.build.extractor.ci.Issues;
-import org.jfrog.build.extractor.ci.MatrixParameter;
-import org.jfrog.build.extractor.ci.Vcs;
 import org.jfrog.build.api.release.Promotion;
+import org.jfrog.build.extractor.builder.BuildInfoMavenBuilder;
+import org.jfrog.build.extractor.ci.*;
 import org.jfrog.build.extractor.clientConfiguration.ArtifactoryClientConfiguration;
 
 import java.io.IOException;
 import java.io.InputStream;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
-import java.util.Arrays;
-import java.util.Date;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
 
-import static org.jfrog.build.api.BuildInfoFields.BUILD_NAME;
-import static org.jfrog.build.api.BuildInfoFields.BUILD_NUMBER;
-import static org.jfrog.build.api.BuildInfoFields.BUILD_STARTED;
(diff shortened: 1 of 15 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -332,4 +332,83 @@ private static void addPropsFromCommandSystemProp(Properties additionalProps, Lo
...
+            return createBuildInfoUrl(url, buildName, buildNumber, timeStamp, project, encode);
+        }
+        if (isNotBlank(project)) {
+            if (endsWith(url, "/artifactory")) {
+                // If the project parameter provided, try to guess the platform URL.
+                return createBuildInfoUrl(removeEnd(url, "/artifactory"), buildName,
+                        buildNumber, timeStamp, project, encode);
+            }
+            // If Artifactory's URL doesn't end with "/artifactory", it is impossible to create the URL.
+            return "";
```
eyalbe4, line 361: Wouldn't it be better to throw an exception here?

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoModelPropertyResolver.java
```diff
@@ -139,10 +125,18 @@ private BuildInfoMavenBuilder resolveCoreProperties(ExecutionEvent event,
...
         String buildStarted = clientConf.info.getBuildStarted();
         if (StringUtils.isBlank(buildStarted)) {
-            buildStarted = new SimpleDateFormat(BuildInfo.STARTED_FORMAT).format(buildStartedDate);
+            buildStarted = DATE_FORMAT.format(buildStartedDate);
+        }
+        long buildMillis = 0;
+        try {
+            buildMillis = DATE_FORMAT.parse(buildStarted).getTime();
+        } catch (ParseException e) {
+            logger.warn(String.format("Artifactory Build Info Model Property Resolver: Couldn't parse start time %s",
```
eyalbe4, line 138: Why don't we throw an exception with this message?

eyalbe4 (changes requested):

build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/client/artifactory/services/PublishBuildInfo.java
```diff
@@ -104,13 +57,16 @@ public HttpRequestBase createRequest() {
...
+        boolean isPlatformUrl = isNotBlank(platformUrl);
+        String url = isPlatformUrl ? platformUrl : client.getUrl();
+        String buildInfoUrl = createBuildInfoUrl(url, build.getName(), build.getNumber(),
+                String.valueOf(build.getStartedMillis()), build.getProject(), true, isPlatformUrl);
+        if (isNotBlank(buildInfoUrl)) {
+            log.info("Build-info successfully deployed. Browse it in Artifactory under " + buildInfoUrl);
         } else {
-            url = createBuildInfoUrl(client.getUrl(), build.getName(), build.getNumber(), true);
+            log.debug("Couldn't create the build-info URL from Artifactory URL: " + client.getUrl());
+            log.info("Build-info successfully deployed.");
```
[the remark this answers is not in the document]
yahavi: No. createBuildInfoUrl may fail to resolve the build info URL in the following specific scenario:
The input URL is Artifactory's URL (not platform) + there is a project + the Artifactory URL does not end with `/artifactory`.
In that case, the build info was published successfully, but it is impossible to guess what is the build-info URL.

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
yahavi: No. This is part of a perfectly legal flow. The build info URL can't be determined in this very specific case. In that case, we may prefer to not display it.
Also, this function is in use in many places and we may not like to add a try-catch mechanism for each one of them.

PR #622, opened 2022-02-17
Filter out unused dependencies in Go builds

yahavi:
-----

Run `go list -f {{with .Module}}{{.Path}} {{.Version}}{{end}} all` to filter unused dependencies in Go builds.
Most of the code copied from https://github.com/jfrog/ide-plugins-common/blob/1.11.0/src/main/java/com/jfrog/ide/common/go/GoTreeBuilder.java#L103. After the release we should delete duplicated code in ide-plugins-common.

### DIFF
--- a/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoDependencyTree.java
+++ b/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoDependencyTree.java
@@ -1,0 +1,124 @@
+package org.jfrog.build.extractor.go.extractor;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang3.exception.ExceptionUtils;
+import org.jfrog.build.api.util.Log;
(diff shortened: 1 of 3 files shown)


build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoDependencyTree.java
```diff
@@ -0,0 +1,124 @@
...
+    private static boolean hasLoop(DependencyTree node, Log logger) {
+        for (DependencyTree parent = (DependencyTree) node.getParent(); parent != null; parent = (DependencyTree) parent.getParent()) {
+            if (Objects.equals(node.getUserObject(), parent.getUserObject())) {
+                logger.debug("Loop detected in " + node.getUserObject());
+                return true;
+            }
+        }
+        return false;
+    }
+
```
eyalbe4, line 123: Redundant new-line.

eyalbe4 (changes requested):

PR #626, opened 2022-03-09
Bugfix - Limit 'mvn deploy' to work with only install/deploy phases

Or-Geva:
-----
Fixes: https://github.com/jfrog/jfrog-cli/issues/1348

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
@@ -141,7 +141,7 @@ public class BuildInfoRecorder extends AbstractExecutionListener implements Buil
             logger.info("Initializing Artifactory Build-Info Recording");
             buildInfoBuilder = buildInfoModelPropertyResolver.resolveProperties(event, conf);
             deployableArtifactBuilderMap = new ConcurrentHashMap<>();
-            skipDeploy(event);
+            setDeploymentPolicy(event);
 
             if (wrappedListener != null) {
                 wrappedListener.sessionStarted(event);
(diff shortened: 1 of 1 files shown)


build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
```diff
@@ -729,7 +730,16 @@ private void cleanUpModule() {
...
+    private void skipDeploy(ExecutionEvent event) {
+        List<String> goals = event.getSession().getRequest().getGoals();
+        // Override the default Maven deploy behavior with Artifactory.
+        if (goals.contains("deploy")) {
+            event.getSession().getUserProperties().put("maven.deploy.skip", Boolean.TRUE.toString());
+            return;
+        }
+        // Skip the artifact deployment behavior if the goals do not contain install phases.
+        if (!goals.contains("install")) {
+            this.conf.publisher.setBooleanValue(ADD_DEPLOYABLE_ARTIFACTS, false);
```
yahavi, line 742: We may still want the deployable artifacts list to allow late deployment
```suggestion
            conf.publisher.setPublishArtifacts(false);
```

yahavi (changes requested):

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
```diff
@@ -729,7 +730,16 @@ private void cleanUpModule() {
     /**
      * Skip the default maven deploy behaviour.
      */
-    private void skipDefaultDeploy(MavenSession session) {
-        session.getUserProperties().put("maven.deploy.skip", Boolean.TRUE.toString());
+    private void skipDeploy(ExecutionEvent event) {
```
yahavi, line 733: Let's think about a better name to reflect this method operation. For example: setDeploymentPolicy.
Please also update the Javadoc.

PR #634, opened 2022-04-01
Allow ignoring errors in "go mod tidy"

yahavi:
-----

Resolves https://github.com/jfrog/jfrog-idea-plugin/issues/208

Allow adding `-e` parameter to `go mod tidy`.


build-info-extractor-go/src/test/java/org/jfrog/build/extractor/go/GoDriverTest.java
```diff
@@ -53,4 +56,27 @@ public void testListUsedModules() throws IOException {
             FileUtils.deleteDirectory(projectDir);
         }
     }
+
+    @Test
+    public void testTidyListUsedModulesErroneous() throws IOException {
```
eyalbe4, line 61: Let's add a descxription for this test method, and also a description for the above ```testTidyListUsedModules``` method.

eyalbe4 (changes requested):

PR #653, opened 2022-06-13
Update dependencies, fix build

elevenfive:
- [X] All [tests](https://ci.appveyor.com/project/jfrog-ecosystem/build-info) passed. If this feature is not already covered by the tests, I added new tests.
-----

Have run all the tests that don't require a docker instance running.  If someone else has one handy and can test this, it would be appreciated!

Fixes #652

PR #659, opened 2022-07-04
Update dependencies, fix build

elevenfive:
-----

Replacement for #653 cc @yahavi


yahavi:
Thanks for this contribution, @elevenfive!
One of the Gradle tests is failing:
>
Gradle suite > Gradle test > org.jfrog.gradle.plugin.artifactory.GradlePluginTest.publicationsTestKotlinDsl[0](4.10.3) FAILED
    org.gradle.testkit.runner.UnexpectedBuildFailure: Unexpected build execution failure in /var/folders/dn/3smrq2yj2qddjccqlky47mpw0000gq/T/gradle_tests_space with arguments [clean, artifactoryPublish, --stacktrace]

    Output:

    FAILURE: Build failed with an exception.

    * What went wrong:
    A problem occurred configuring root project 'gradle_tests_space'.
    > Could not open cache directory 55ogsh56r3ls2qvvelop96jkt (/private/var/folders/dn/3smrq2yj2qddjccqlky47mpw0000gq/T/.gradle-test-kit-yahavi/caches/4.10.3/gradle-kotlin-dsl-accessors/55ogsh56r3ls2qvvelop96jkt).
       > org.jetbrains.kotlin.protobuf.InvalidProtocolBufferException: Protocol message contained an invalid tag (zero).

Steps to reproduce:
* Run `./gradlew build-info-extractor-gradle:test`

Would you consider taking a look?

elevenfive:
Hi @yahavi, thank you for looking!   I just updated my branch to resolve conflicts with main branch and also to pull new updates as applicable.

- We can upgrade to Java 11, I'm not sure if the project is ready for that (builder env must have Java 11 of course), so I put notes for that regarding updating the Testng library

- All tests that don't require network connection to the test instance pass for me, can you retest with the updates I posted?

yahavi:
Thanks @elevenfive,
1. This library is a dependency for many old clients and we prefer migrating to Java 11 only if there is no other choice. We had some issues after the migration from Java 7 to Java 8. AFAIK Java 8 should receive security updates until 2030.
2. Unfortunately `publicationsTestKotlinDsl` failed again. 
You can start an OSS server to run the Gradle tests. See https://github.com/jfrog/build-info#testing-on-artifactory-oss. 
I don't know what is the root cause, but it looks like upgrading Gradle from 5 to 6 caused this issue. I'm not sure whether Gradle 4 with Kotlin is in use by somebody, maybe we can drop Gradle 4 support on the Gradle plugin and skip the test on Gradle 4 - @eyalbe4 WDYT?

elevenfive:
Thank you I will try to repro locally and see if there is a fix.

AFAICT Gradle 4 & 5 are both EOL, with 5 last receiving an update in 2019 and 4 last receiving an update in 2018; I would personally suggest dropping Gradle 4 support.

https://gradle.org/releases/

yahavi:
Hi @elevenfive,
We had a team conversation about it. Dropping Gradle 4's support requires a new major (5).
We do have plans in the near future to separate the Gradle Artifactory plugin code from the build-info repository and release a new major. After the separation, it will be easier to make structural changes and dependency upgrades in the build-info repository.
In the meanwhile, we prefer to use Gradle 5 to build this project.

elevenfive:
Gotcha - I have reverted back to Gradle 5 and added a comment.  Thanks for looking into it team!

yahavi (approved):
It looks good now.
Thanks for your contribution, @elevenfive!

PR #677, opened 2022-11-23
CommandExecutor - Support executable path with spaces

yahavi:
-----

Resolve an issue in the JFrog IDEA plugin whereby Go executable path is under `C:\program files` directory in Windows.

### DIFF
--- a/build-info-extractor/src/test/java/org/jfrog/build/extractor/executor/CommandExecutorTest.java
+++ b/build-info-extractor/src/test/java/org/jfrog/build/extractor/executor/CommandExecutorTest.java
@@ -76,10 +76,11 @@ public class CommandExecutorTest {
             assertTrue(executableFile.setExecutable(true));
             FileUtils.writeStringToFile(executableFile, "echo \"I have spoken\"", StandardCharsets.UTF_8);
 
+            // Run executable in "Kuiil space/ship.bat" with no arguments and make sure it prints "I have spoken"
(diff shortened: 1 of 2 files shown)


build-info-extractor/src/test/java/org/jfrog/build/extractor/executor/CommandExecutorTest.java
```diff
@@ -61,4 +65,20 @@ public void testExeCommand() {
...
+    @Test
+    public void testExeCommandWithSpaces() throws IOException, InterruptedException {
+        Path tmpDir = Files.createTempDirectory("testEscapeSpacesInPath");
+        try {
+            Path testDir = Files.createDirectories(tmpDir.resolve("Kuiil space"));
+            File executableFile = Files.createFile(testDir.resolve("ship.bat")).toFile();
+            assertTrue(executableFile.setExecutable(true));
+
+            CommandExecutor commandExecutor = new CommandExecutor(executableFile.getAbsolutePath(), null);
+            CommandResults results = commandExecutor.exeCommand(null, new ArrayList<>(), new ArrayList<>(), null);
```
eyalbe4, line 78: It appears that we're running here an empty command. I suggest we actually try run something here, like a shell sceipt that echo's "hello" and then check the result.

eyalbe4 (changes requested):

PR #678, opened 2022-11-23
Set the connection request timeout

yahavi:
-----

Set the connection request timeout to avoid infinite timeout during connection requests.

### DIFF
--- a/build-info-client/src/main/java/org/jfrog/build/client/PreemptiveHttpClientBuilder.java
+++ b/build-info-client/src/main/java/org/jfrog/build/client/PreemptiveHttpClientBuilder.java
@@ -181,6 +181,7 @@ public class PreemptiveHttpClientBuilder {
                 .custom()
                 .setSocketTimeout(timeoutMilliSeconds)
                 .setConnectTimeout(timeoutMilliSeconds)
+                .setConnectionRequestTimeout(timeoutMilliSeconds)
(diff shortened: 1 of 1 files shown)


build-info-client/src/main/java/org/jfrog/build/client/PreemptiveHttpClientBuilder.java
```diff
@@ -181,6 +181,7 @@ protected HttpClientBuilder createHttpClientBuilder() {
                 .custom()
                 .setSocketTimeout(timeoutMilliSeconds)
                 .setConnectTimeout(timeoutMilliSeconds)
+                .setConnectionRequestTimeout(timeoutMilliSeconds)
```
eyalbe4, line 184: It looks like the value we set for `timeoutMilliSeconds` is one second? Is it not?
This seems like a very short timeframe. I suggest changing it to 30 seconds, and apply it to 3 timeout values.

eyalbe4 (changes requested):

build-info-client/src/main/java/org/jfrog/build/client/PreemptiveHttpClientBuilder.java
yahavi: It is not 1 second, the default is 5 minutes

PR #697, opened 2023-02-04
Only the deploy goal should upload artifacts

markrekveld:
The build info recorder will upload artifacts if either the deploy or install goals are used, but deployments should only be done when the deploy goal is used.

Using install to deploy artifacts to Artifactory results in unexpected behaviour and can lead to build failures and artifact corruption.

This PR restores the deploy behaviour to the expected goals of Maven.

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
@@ -708,7 +708,7 @@ public class BuildInfoRecorder extends AbstractExecutionListener implements Buil
     }
 
     /**
-     * Only Maven deploy / install goals are supposed to upload artifacts to artifactory.
+     * Only Maven deploy goal are supposed to upload artifacts to artifactory.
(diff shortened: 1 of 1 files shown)


yahavi:
Thanks for your contribution, @markrekveld!
I'm afraid that this PR introduces a significant breaking change. It may require releasing major versions of some of our plugins.
For now, we have some workarounds for each one of them. 
For example, if you're using the Jenkins Artifactory plugin, you can do the following:

Scripted pipeline:
```groovy
rtMaven.deployer.deployArtifacts = false
```

Declarative pipeline:
```groovy
rtMavenDeployer (
    deployArtifacts: false
)
```

Please let me know if that helped.

markrekveld:
I understand this small but significant change may require some more thought.

Can you share the workaround for using the JFrog CLI on JFrog Pipelines?
Currently I don't configure the deployer server id and repos but would like a better solution if it exists.

yahavi:
Sure, @markrekveld.
To prevent deployment, add the `artifactory.publish.artifacts=false` system property.

For example:
```yml
mvnCommand: clean install -Dartifactory.publish.artifacts=false
```

Please let me know if that helped.

PR #707, opened 2023-02-16
Refactor and handle unhandled exceptions

omerzi:
-----

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
@@ -21,6 +21,7 @@ import org.jfrog.build.extractor.clientConfiguration.PatternMatcher;
 
 import java.io.*;
 import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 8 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -179,36 +180,37 @@ public static String buildInfoToJsonString(BuildInfo buildInfo) throws IOExcepti
         JsonFactory jsonFactory = createJsonFactory();
 
         StringWriter writer = new StringWriter();
-        JsonGenerator jsonGenerator = jsonFactory.createJsonGenerator(writer);
-        jsonGenerator.useDefaultPrettyPrinter();
-
-        jsonGenerator.writeObject(buildInfo);
-        String result = writer.getBuffer().toString();
-        return result;
+        try (JsonGenerator jsonGenerator = jsonFactory.createGenerator(writer)) {
```
yahavi, line 183: Let's add the StringWriter to the try..with..resources

yahavi (changes requested):

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
omerzi: done

PR #716, opened 2023-03-02
Filter build-info and modules properties

Or-Geva:
-----
Prior to this PR, build-info properties were filtered only for environment variables and system properties.
The PR moves all filter logic into one place and performs exclude/include logic for all properties, including:

1. Environment variables (including properties with `buildInfo.env` prefix)
2. System properties
3. Properties from file
4. Build-info modules properties
5. Exclude not just the key but also the value of properties

### DIFF
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
@@ -178,7 +178,6 @@ public class GradleBuildInfoExtractor implements BuildInfoExtractor<Project> {
 
         BuildInfo buildInfo = bib.build();
         PackageManagerUtils.collectEnvIfNeeded(clientConf, buildInfo);
-        PackageManagerUtils.filterBuildInfoProperties(clientConf, buildInfo, clientConf.getLog());
         log.debug("buildInfoBuilder = " + buildInfo);
 
         return buildInfo;
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
@@ -176,13 +176,11 @@ public class GradleBuildInfoExtractor implements BuildInfoExtractor<Project> {
             bib.addRunParameters(matrixParameter);
         }
 
-        log.debug("buildInfoBuilder = " + bib);
-        // for backward compatibility for Artifactory 2.2.3
         BuildInfo buildInfo = bib.build();
-        if (parentName != null && parentNumber != null) {
-            buildInfo.setParentBuildId(parentName);
-        }
         PackageManagerUtils.collectEnvIfNeeded(clientConf, buildInfo);
+        PackageManagerUtils.filterBuildInfoProperties(clientConf, buildInfo);
+        log.debug("buildInfoBuilder = " + buildInfo);
+
         return buildInfo;
     }
 }
--- a/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
+++ b/build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
@@ -177,7 +177,7 @@ public class GradleBuildInfoExtractor implements BuildInfoExtractor<Project> {
         }
 
         BuildInfo buildInfo = bib.build();
-        PackageManagerUtils.collectEnvIfNeeded(clientConf, buildInfo);
+        PackageManagerUtils.collectAndFilterEnvIfNeeded(clientConf, buildInfo);
         log.debug("buildInfoBuilder = " + buildInfo);
 
         return buildInfo;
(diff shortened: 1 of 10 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerUtils.java
```diff
@@ -1,19 +1,31 @@
...
 import java.util.Properties;
 
 /**
  * Created by Bar Belity on 12/07/2020.
  */
 public class PackageManagerUtils {
+    private static final String apiKeySecretPrefix = "AKCp8";
+    private static final int apiKeySecretMinimalLength = 73;
+    private static final String referenceTokenSecretPrefix = "cmVmdGtuOjAxOj";
+    private static final int referenceTokenSecretMinimalLength = 64;
```
yahavi, line 26: I almost sure that reference tokens are 64 characters fixed

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -107,16 +114,10 @@ public static Properties stripPrefixFromProperties(Properties source, String pre
     }
 
     public static Properties getEnvProperties(Properties startProps, Log log) {
-        IncludeExcludePatterns patterns = new IncludeExcludePatterns(
```
yahavi, line 110: Removing the filtering code from this method sounds very risky to me. Let's do the "get" and the "filter" logic in one place to enforce filtering in all flows.

yahavi (changes requested):
Please don't merge this code before my final approval. I want to go through it again.

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
```diff
@@ -176,13 +176,11 @@ public BuildInfo extract(Project rootProject) {
             bib.addRunParameters(matrixParameter);
         }
 
-        log.debug("buildInfoBuilder = " + bib);
-        // for backward compatibility for Artifactory 2.2.3
         BuildInfo buildInfo = bib.build();
-        if (parentName != null && parentNumber != null) {
-            buildInfo.setParentBuildId(parentName);
-        }
```
yahavi, line 184: Why is that?

build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerUtils.java
```diff
@@ -65,4 +76,70 @@ public static void collectEnvIfNeeded(ArtifactoryClientConfiguration clientConfi
...
+     * Done by searching for a known constant prefix of the secret and verifying the length of the substring is sufficient to include the expected length of the secret.
+     *
+     * @param variableValue       - string to search in
+     * @param secretPrefix        - secret constant prefix
+     * @param secretMinimalLength - secret minimal expected length
+     * @return whether a secret is suspected
+     */
+    private static boolean containsSuspectedSecret(String variableValue, String secretPrefix, int secretMinimalLength) {
+        int secretIndex = variableValue.indexOf(secretPrefix);
+        return secretIndex > -1 && variableValue.substring(secretIndex).length() >= secretMinimalLength;
```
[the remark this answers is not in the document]
Or-Geva: Changed to :
```
return variableValue.startsWith(secretPrefix) && variableValue.length() >= secretMinimalLength;
```

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleBuildInfoExtractor.java
Or-Geva: Looks like this is related to:
```
for backward compatibility for Artifactory 2.2.3
```

build-info-extractor/src/test/java/org/jfrog/build/extractor/packageManager/PackageManagerUtilsTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+import java.util.Date;
+import java.util.Properties;
+
+import static org.jfrog.build.extractor.BuildInfoExtractorUtils.getEnvProperties;
+import static org.jfrog.build.extractor.packageManager.PackageManagerUtils.filterBuildInfoProperties;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.assertNull;
+
+public class PackageManagerUtilsTest {
```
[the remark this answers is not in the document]
Or-Geva: Sure, it is already tested [here](https://github.com/jfrog/build-info/pull/716/files/c382d3faa3423cf8044e7452b48621574fdd0e64#diff-c907117809abd7afd5787cd69de06463dde16bdc75877599532a497f57956e38R97)

build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
Or-Geva: I understand the concern, but this is not the place to do filtering. According to the name of this function, it returns all environment variables. By moving the filter logic to a new location, we can filter not only the environment variables, but also modules. Putting the filter logic here would prevent us from filtering things unrelated to environment variables, so we would have to split the logic across the code, and that will cause a mess in the code.

Or-Geva: Due to security risks, we will return the filter logic to here despite the above comment .

yahavi (changes requested):

build-info-extractor/src/main/java/org/jfrog/build/extractor/packageManager/PackageManagerUtils.java
```diff
@@ -64,5 +76,73 @@ public static void collectEnvIfNeeded(ArtifactoryClientConfiguration clientConfi
             return;
         }
         buildInfo.setProperties(envProperties);
+        PackageManagerUtils.filterBuildInfoProperties(clientConfiguration, buildInfo, clientConfiguration.getLog());
```
yahavi, line 79:
```suggestion
       filterBuildInfoProperties(clientConfiguration, buildInfo, clientConfiguration.getLog());
```

PR #717, opened 2023-03-08
Add Client ID to the usage report

talarian1:
-----
Add client ID to the usage report - we use the first network interface MAC address we found as the client ID.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
@@ -16,20 +16,16 @@ import java.util.Enumeration;
 public class UsageReporter {
     private final String productId;
     private FeatureId[] features;
-    private final String uniqueClientId;
+    private String uniqueClientId;
 
     public UsageReporter(String productId, String[] featureIds) {
         this.productId = productId;
         setFeatures(featureIds);
-        this.uniqueClientId = generateUniqueClientId();
     }
 
-    private String generateUniqueClientId() {
+    private String generateUniqueClientId() throws SocketException {
         byte[] macAddress = getMacAddress();
-        if (macAddress != null) {
-            return new String(DigestUtils.sha1(macAddress));
-        }
-        return null;
+        return macAddress != null ? DigestUtils.sha1Hex(macAddress) : null;
     }
 
     @SuppressWarnings("unused")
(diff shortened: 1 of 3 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
```diff
@@ -1,40 +1,63 @@
...
+    @SuppressWarnings("unused")
     public void reportUsage(String artifactoryUrl, String username, String password, String accessToken, ProxyConfiguration proxyConfiguration, Log log) throws IOException {
         try (ArtifactoryManager artifactoryManager = new ArtifactoryManager(artifactoryUrl, username, password, accessToken, log)) {
             if (proxyConfiguration != null) {
                 artifactoryManager.setProxyConfiguration(proxyConfiguration);
             }
+            try {
+                uniqueClientId = uniqueClientId == null ? generateUniqueClientId() : uniqueClientId;
+            } catch (SocketException e) {
+                log.debug("Wasn't able to generate unique client ID.");
```
yahavi, line 40: Exceptions mustn't be ignored
```suggestion
                log.warn("Wasn't able to generate unique client ID: " + ExceptionUtils.getRootCauseMessage(e));
```

build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
```diff
@@ -1,40 +1,63 @@
...
+    private final String productId;
     private FeatureId[] features;
+    private String uniqueClientId;
 
     public UsageReporter(String productId, String[] featureIds) {
         this.productId = productId;
         setFeatures(featureIds);
     }
 
+    private String generateUniqueClientId() throws SocketException {
```
yahavi, line 26: Let's add some JavaDoc to this function. Let's mention that the original MAC address is impossible to retrieve because we are using sha1 hashing.

build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
```diff
@@ -1,40 +1,63 @@
...
     public void reportUsage(String artifactoryUrl, String username, String password, String accessToken, ProxyConfiguration proxyConfiguration, Log log) throws IOException {
         try (ArtifactoryManager artifactoryManager = new ArtifactoryManager(artifactoryUrl, username, password, accessToken, log)) {
             if (proxyConfiguration != null) {
                 artifactoryManager.setProxyConfiguration(proxyConfiguration);
             }
+            try {
+                uniqueClientId = uniqueClientId == null ? generateUniqueClientId() : uniqueClientId;
+            } catch (SocketException e) {
+                log.debug("Wasn't able to generate unique client ID.");
+            }
```
yahavi, line 41: Shouldn't the code that populates the class variable "uniqueClientId" be in the constructor?

build-info-extractor/src/main/java/org/jfrog/build/extractor/usageReport/UsageReporter.java
yahavi, line 26: Let's move the private function below "reportUsage" to make this class more readable

yahavi (changes requested):

yahavi (approved):

PR #739, opened 2023-06-18
Add timeout for exeCommand utility

talarian1:
-----

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
@@ -22,8 +22,9 @@ import static java.lang.String.join;
 public class CommandExecutor implements Serializable {
     private static final long serialVersionUID = 1L;
     private static final int TIMEOUT_EXIT_VALUE = 124;
-    private static final int TIMEOUT_SECONDS = 30;
-    private static final int TIMEOUT_MINUTES = 120;
+    private static final int READER_SHUTDOWN_TIMEOUT_SECONDS = 30;
+    private static final int PROCESS_TERMINATION_TIMEOUT_SECONDS = 30;
+    private static final int EXECUTION_TIMEOUT_MINUTES = 120;
     private final String[] env;
     private final String executablePath;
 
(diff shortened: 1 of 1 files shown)


asafgabai (changes requested):

build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -154,8 +180,7 @@ public CommandResults exeCommand(File execDir, List<String> args, List<String> c
     private CommandResults getCommandResults(boolean terminatedProperly, List<String> args, String output, String error, int exitValue) {
         CommandResults commandRes = new CommandResults();
         if (!terminatedProperly) {
-            error += System.lineSeparator() + format("Process '%s' had been terminated forcibly after timeout of %d seconds.",
-                    join(" ", args), TIMEOUT_SECONDS);
+            error += System.lineSeparator() + format("Process '%s' had been terminated forcibly after timeout of %d minutes.", join(" ", args), EXECUTION_TIMEOUT_MINUTES);
```
asafgabai, line 183: If the timeout is overridden by a parameter, then this log message will be inaccurate. It'd be best to use the timeout parameter instead of the const here.

build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -123,28 +124,53 @@ static String maskCredentials(String command, List<String> credentials) {
...
+                // Ensure termination of the subprocess we have created.
+                if (process.isAlive()) {
+                    // First try to terminate the process gracefully
+                    process.destroy();
+                    process.waitFor(PROCESS_TERMINATION_TIMEOUT_SECONDS, TimeUnit.SECONDS);
+                    if (process.isAlive()) {
+                        // Finally, force kill.
+                        process.destroyForcibly();
+                    }
+                }
```
asafgabai, line 173: Instead of closing the readers before the process termination, it may be better to close them after the process finishes. This way, we ensure that all output generated by the process is captured, including any output that might occur following its termination.
What do you think?

asafgabai (changes requested):

PR #748, opened 2023-07-16
Gradle legacy Archives are not published

yahavi:
-----

When the Gradle Java plugin is not applied, Legacy Gradle Archive configurations are not automatically collected and published in CI mode.


build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/listener/ProjectsEvaluatedBuildListener.groovy
```diff
@@ -231,7 +234,7 @@ public class ProjectsEvaluatedBuildListener extends BuildAdapter implements Proj
 
             // Add publications to Artifactory task
             artifactoryTask.addDefaultPublications()
-        } else {
+        } else if (artifactoryTask.project.plugins.hasPlugin("maven")) {
```
attiasas, line 237: please insert a comment to explain why we need this plugin for configuration

attiasas (changes requested):

PR #749, opened 2023-07-19
Fix get build number of AQL

yahavi:
-----

Fix an issue caused by https://github.com/jfrog/build-info/pull/740.
In AQL queries for builds, we always try to get the latest build number, regardless of its value.
We should try to get the latest build number only in one of the following conditions:
1. buildNumber is "LATEST"
2. buildNumber is "LAST_RELEASE"
3. buildNumber is empty

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/AqlHelper.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/AqlHelper.java
@@ -32,9 +32,9 @@ public class AqlHelper {
     }
 
     private void buildQueryAdditionalParts() throws IOException {
-        this.buildName = getBuildName(this.filesGroup.getBuild());
+        this.buildName = getBuildName();
         // AQL doesn't support projects.
-        this.buildNumber = getBuildNumber(artifactoryManager, buildName, this.filesGroup.getBuild(), null);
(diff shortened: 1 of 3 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/clientConfiguration/util/AqlHelper.java
```diff
@@ -133,15 +135,28 @@ private String getBuildNumber(ArtifactoryManager client, String buildName, Strin
...
+            if (isBuildLatestType(buildNumber)) {
+                String retrievedBuildNumber = artifactoryManager.getLatestBuildNumber(buildName, buildNumber, null);
+                if (retrievedBuildNumber == null) {
+                    logBuildNotFound(buildName, buildNumber);
+                }
+                buildNumber = retrievedBuildNumber;
             }
-            buildNumber = retrievedBuildNumber;
         }
         return buildNumber;
```
Or-Geva, line 146:
```java
if (!isBuildLatestType(buildNumber)) {
    return buildNumber;
}
String retrievedBuildNumber = artifactoryManager.getLatestBuildNumber(buildName, buildNumber, null);
if (StringUtils.isBlank(retrievedBuildNumber)) {
    logBuildNotFound(buildName, buildNumber);
}
return retrievedBuildNumber;
```

PR #757, opened 2023-09-18
Improve handling of spaces in command execution utility on Windows

talarian1:
-----


build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -204,7 +194,37 @@ private static Process runProcess(File execDir, List<String> args, List<String>
...
+        return executablePath.trim().replaceAll(" ", SystemUtils.IS_OS_WINDOWS ? "^ " : "\\\\ ");
+    }
+
+    /**
+     * Removes the surrounding quotes from the given string, if present.
+     *
+     * @param str The string to unquote.
+     * @return The unquoted string.
+     */
+    public static String unQuote(String str) {
```
asafgabai, line 223: The word "unquote" has a meaning different from the one used here, so I suggest renaming this method.
Suggestion: `removeQuotes`

asafgabai (changes requested):

PR #777, opened 2024-01-04
Fix running executables with spaces in their paths in Windows

asafgabai:
-----
When executing a command on Windows, add its directory path to the Path environment variable and execute it using its name. We do that to avoid passing an executable path that contains spaces, which causes parsing issues in Windows CMD.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
@@ -36,7 +36,7 @@ public class CommandExecutor implements Serializable {
      * @param env            - Environment variables to use during execution.
      */
     public CommandExecutor(String executablePath, Map<String, String> env) {
-        this.executablePath = escapeSpacesInPath(executablePath);
+        this.executablePath = executablePath.trim();
         Map<String, String> finalEnvMap = new HashMap<>(System.getenv());
         if (env != null) {
             Map<String, String> fixedEnvMap = new HashMap<>(env);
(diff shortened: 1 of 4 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/executor/CommandExecutor.java
```diff
@@ -179,13 +180,21 @@ private CommandResults getCommandResults(boolean terminatedProperly, List<String
...
             args.addAll(credentials);
         }
         if (SystemUtils.IS_OS_WINDOWS) {
+            Path execPath = Paths.get(executablePath);
+            if (execPath.isAbsolute()) {
+                env = generateWindowsEnv(execPath, env);
+                args.add(0, execPath.getFileName().toString());
+            } else {
+                args.add(0, executablePath.replaceAll(" ", "^ "));
+            }
```
Or-Geva, line 194: We can break down the functionality into smaller, well-named methods.
```
private static void formatProcess(List<String> args, List<String> credentials, String executablePath, Map<String, String> env) {
    if (credentials != null) {
        args.addAll(credentials);
    }

    if (SystemUtils.IS_OS_WINDOWS) {
        formatWindowsProcess(executablePath, args, env);
    } else {
        formatUnixProcess(executablePath, args);
    }
}

private static void formatWindowsProcess(String executablePath, List<String> args, Map<String, String> env) {
    Path execPath = Paths.get(executablePath);
    if (execPath.isAbsolute()) {
        env = generateWindowsEnv(execPath, env);
        args.add(0, execPath.getFileName().toString());
    } else {
        args.add(0, executablePath.replaceAll(" ", "^ "));
    }
    args.addAll(0, Arrays.asList("cmd", "/c"));
}

private static void formatUnixProcess(String executablePath, List<String> args) {
    args.add(0, executablePath.replaceAll(" ", "\\\\ "));
    String strArgs = join(" ", args);
    args = new ArrayList<String>() {{
        add("/bin/sh");
        add("-c");
        add(strArgs);
    }};
}
```

Or-Geva (changes requested):

PR #787, opened 2024-03-21
Correct .npmrc for newer versions of NPM

Spaction:
-----
Replaces old _auth from the temp npmrc file conditionally if we are past or equal the version 8.19 which introduced the new format.

### DIFF
--- a/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
+++ b/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
@@ -97,6 +97,32 @@ public class NpmDriver implements Serializable {
         }
     }
 
+    public int compareVersionTo(File workingDirectory, String compareVersion){
+        String currentVersion = this.version(workingDirectory);
+
+        List<Integer> currentComponents = Arrays.stream(currentVersion.split("\\."))
+                                                .map(Integer::parseInt)
+                                                .collect(Collectors.toList());
+        List<Integer> compareComponents = Arrays.stream(compareVersion.split("\\."))
+                                                .map(Integer::parseInt)
+                                                .collect(Collectors.toList());
+
+        int currentSize = currentComponents.size();
+        int compareSize = compareComponents.size();
+        int maxLength = Math.max(currentSize, compareSize);
+        
+        for (int i = 0; i < maxLength; i++){
+            int currentComp = i < currentSize ? currentComponents.get(i) : 0;
+            int compareComp = i < compareSize ? compareComponents.get(i) : 0;
+
+            int comp = compareComp - currentComp;
+            if(comp != 0){
+                return comp;
+            }
+        }
+        return 0;
+    }
+
     public String version(File workingDirectory) throws IOException, InterruptedException {
         return runCommand(workingDirectory, new String[]{"--version"}, Collections.emptyList()).getRes();
     }
--- a/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
+++ b/build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
@@ -97,8 +97,8 @@ public class NpmDriver implements Serializable {
         }
     }
 
-    public int compareVersionTo(File workingDirectory, String compareVersion){
-        String currentVersion = this.version(workingDirectory);
+    public int compareVersionTo(File workingDirectory, String compareVersion) throws IOException, InterruptedException {
+        String currentVersion = version(workingDirectory);
 
         List<Integer> currentComponents = Arrays.stream(currentVersion.split("\\."))
                                                 .map(Integer::parseInt)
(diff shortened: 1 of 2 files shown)


Spaction:

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
```diff
@@ -189,9 +189,24 @@ private void createTempNpmrc(Path workingDir, List<String> commandArgs) throws I
...
+                String newAuthKey = artifactoryManager.getUrl();
+                if (!StringUtils.endsWith(newAuthKey, "/")) {
+                    newAuthKey += "/";
+                }
+                newAuthKey += "api/npm/:_auth";
+                newAuthKey = newAuthKey.replaceAll("^http(s)?:","") + ":_auth";
+                npmAuth.setProperty(newAuthKey, authProp);
+                npmAuth.remove("_auth");
+            }
+        }
```
eyalbe4, line 206: * To make the code easier to read. let move the above code into a new privated method named `handleNpmCompatibility`
* Our coding conversion include having a single white-space after //
* To align with our coding conversions, the following if statement - 
`if( this.npmDriver.compareVersionTo(workingDir.toFile(), "8.19") >= 0){`
should be modified to - 
`if (this.npmDriver.compareVersionTo(workingDir.toFile(), "8.19") >= 0) {`
(notice the white-spacing conventions)
* Let's add a unit test for this functionality in `NpmBuildInfoExtractorTest`.

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/NpmDriver.java
```diff
@@ -97,6 +97,32 @@ public JsonNode list(File workingDirectory, List<String> extraArgs) throws IOExc
         }
     }
 
+    public int compareVersionTo(File workingDirectory, String compareVersion) throws IOException, InterruptedException {
```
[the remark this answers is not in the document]
Spaction: This is validating that NPM is of a specific version, not Artifactory.

build-info-extractor-npm/src/main/java/org/jfrog/build/extractor/npm/extractor/NpmBuildInfoExtractor.java
Spaction: The added test case is 99% the same as the NpmExtractorTest.npmCiTest, which under the NpmInstallCi and subsequent .execute does in-fact call the NPMBuildInfoExtractor method. There is nothing specific for the test to check except that it should pass for both the Node 15 and Node 20 now found in the integrationTests.yml. I'm unable to execute this fully on my side as I do not have Admin access to an Artifactory system and the open source does not have NPM enabled.

Spaction:
I'm not part of the dev-ops team, but i'll mention to them that this Plug-in is not being as actively developed.  As mentioned, the part where this should shine is with running against Node 15 and Node 20, where as before that would have failed.

Spaction:
@eyalbe4  Any chance you can re-run the checks?

Spaction:
@eyalbe4  Is there anything else needing to be done for this to be rechecked?

Spaction:
@yahavi Are you able to re-execute the checks?

Spaction (commented):
Updated

RobiNino:
@Spaction , Sorry for the delay, we'll review this soon

RobiNino:
Hi @Spaction ,
I took the liberty to update the PR with a few fixes to the tests infrastructure. I also temporarily changed the target branch so that the changes to the workflow made in this PR will reflect in the tests.

Seems like the [npm tests are failing](https://github.com/jfrog/build-info/actions/runs/9195040145/job/25289927938). Could you please look into it before we proceed?
Thanks

Spaction:
@RobiNino,

Updated the Unit tests. Can honestly say the test doesn't differ between between the NPMExtractor Test vs the one for NpmBuildInfo. The expanded test for the v15 vs v20 is what would have failed previously during the NPMExtractorTest. Only way to confirm this would be to expand only the unit test to include Node v20 and see that it fails for that branch. This is only to say that the new test can be excluded as the other indirectly tests the method\execution path in its npmCiTest

Spaction:
@RobiNino Any updates on my last resposne?

Spaction:
@RobiNino , @eyalbe4  Any update?

Spaction:
@RobiNino , @eyalbe4 ?

eyalbe4:
@Spaction,
While we'd like to continue promoting this, the team is currently loaded with a few other initiatives and will get back to this when they become available. 
As a side note, Artifactory's build info integration with NPM and other package managers is strongly recommend by us through JFrpg CLI and not through this older java based build-info-extractor code. The JFrog CLI integration is already more stable, has better support for NPM and receives a lot more attention from the community and the dev teams.

Spaction:
@eyalbe4 You mentioned that back in March, to which I'll repeat what I said earlier. I've no control over what tools we use. I've brought up to the team responsible to use the newer library and for reasons I won't state here was told it would not be happening.

PR #798, opened 2024-07-17
Replace NettyDockerCmdExecFactory for OkHttpDockerCmdExecFactory

### LINKED ISSUE #797: Add npipe support for build-info-extractor-docker
**Describe the bug**
On windows using windows containers is not possible to use the npipe protocol to publish docker images to the artifactory.
The NettyDockerCmdExecFactory does not support the npipe protocol which has been used here:
https://github.com/jfrog/build-info/blob/master/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerJavaWrapper.java#L77
I found this comment in the docker-java repository:
https://github.com/docker-java/docker-java/issues/765#issuecomment-588942835
So I suggest the replacement from NettyDockerCmdExecFactory  to OkHttpDockerCmdExecFactory in order to support npipe.

**To Reproduce**
Configure windows docker engine to use windows containers.
Configure a Jenkins instance with docker clouds. 
Add the following mount to the docker cloud template (exposes the host docker engine to the ephemeral container):
type=npipe,source=\\.\pipe\docker_engine,destination=\\.\pipe\docker_engine

Add and execute a pipeline that makes use of the Artifactory.docker (see example in the first screenshot).

**Expected behavior**
At red is the affected code leading to the exception shown in the second screenshot.
If we just use the docker-plugin from jenkins there is no issues.

**Screenshots**
image
image

**Versions**
- Extractor version: v2.41.21
- Operating system: Windows 10
- Artifactory Version: 7.84.17

**Additional context**

Ru13en:
Add support to npipe protocol for windows docker by using okHttp.
Fixes #797 

-----

### DIFF
--- a/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerJavaWrapper.java
+++ b/build-info-extractor-docker/src/main/java/org/jfrog/build/extractor/docker/DockerJavaWrapper.java
@@ -10,7 +10,7 @@ import com.github.dockerjava.core.DockerClientConfig;
 import com.github.dockerjava.core.command.BuildImageResultCallback;
 import com.github.dockerjava.core.command.PullImageResultCallback;
 import com.github.dockerjava.core.command.PushImageResultCallback;
-import com.github.dockerjava.netty.NettyDockerCmdExecFactory;
+import com.github.dockerjava.okhttp.OkHttpDockerCmdExecFactory;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.commons.lang3.tuple.Pair;
 import org.jfrog.build.api.util.Log;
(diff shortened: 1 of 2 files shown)


Ru13en:

Ru13en:
recheck

eyalbe4:
@Ru13en,
For integrating Jenkins with the JFrog Platform, our recommendation is to avoid using the old Artifactory Jenkins Plugin, and use the newer [JFrog Jenkins Plugin](https://plugins.jenkins.io/jfrog/) whenever it is possible. The older plugin, because of its design, introduces multiple issues, especially when it comes to the integration with docker. We will therefore refrain from making new investments in the Artifactory Jenkins Plugin.

Ru13en:
Hi @eyalbe4 , thank you for the info. I will give it a try then :). BR

PR #812, opened 2024-12-16
Replace netty with Apache httpclient

### LINKED ISSUE #615: Docker build info extractor does not work on arm64 Linux
Hello,

recently I wanted to use the JFrog Jenkins Artifactory plugin on an arm64 machine to push docker images. That badly failed because this `build-info` package relies on `docker-java` which does not embed/depend on the arm64 Linux shared library for Netty.
I opened an upstream issue at docker-java/docker-java#1743 but the author declined and stated that consumers easily could add that dependency on their own or just stop relying on the deprecated Netty transport.

Would you be able to switch the transport from Netty to Apache HTTP Client or add an extra dependency on Netty libraries for `linux-aarch_64` epoll (like the PR above tried to)?

Thanks,
Gregor

jkenn99:
-----

The Netty transport is no longer supported in docker-java. This PR replaces it with the Apache HttpClient transport. See https://github.com/docker-java/docker-java/blob/main/docs/transports.md

Fixes #615.


jkenn99:

PR #815, opened 2025-01-14
Fix go version extractor to accept valid semver with 2 dashes

nadav-y:
-----

### DIFF
--- a/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoVersionUtils.java
+++ b/build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoVersionUtils.java
@@ -1,11 +1,11 @@
 package org.jfrog.build.extractor.go.extractor;
 
+import com.github.zafarkhaja.semver.Version;
 import org.apache.commons.lang3.StringUtils;
 import org.jfrog.build.api.util.Log;
 
 import java.io.File;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
+import java.util.Optional;
 
 /**
  * @author BarakH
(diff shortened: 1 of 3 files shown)


nadav-y:

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoVersionUtils.java
```diff
@@ -14,30 +14,24 @@ public class GoVersionUtils {
...
-                    return Integer.parseInt(major);
-                } catch (NumberFormatException e) {
-                    log.error("Failed to parse major version of " + version, e);
-                }
-            }
+        Optional<Version> parsedVersion = Version.tryParse(StringUtils.removeStart(version, "v"));
+        if (parsedVersion.isPresent()) {
+            return parsedVersion.get().majorVersion();
+        } else {
+            log.debug("Failed to parse major version of " + version);
```
RobiNino, line 34: We used to log an error if we failed to parse, I suggest we stick to the same behavior. WDYT?

RobiNino (changes requested):

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoVersionUtils.java
nadav-y: we logged only if it was a valid "semver" by our rules/regex - and only the first part wasn't a digit (which shouldn't be possible).
On non-valid version (!atcher.matches()) we logged nothing.
This (parsedVersion.isPresent()) is the equivalent to that- so logging and error here would change the previous behavior

RobiNino: Agreed, thanks

RobiNino (approved):

PR #823, opened 2025-06-11
Fix for Jenkins Malformed encoding

agrasth:
-----

The issue was that when loading properties files with malformed Unicode escape sequences (like `\uxxxx`), the original code would fail with `IllegalArgumentException: Malformed \uxxxx encoding`.
The fix implements a fallback strategy:
- First, try the normal properties loading
- If that fails with `IllegalArgumentException`, try loading with explicit UTF-8 encoding
- If that also fails, continue with empty properties rather than failing the entire build

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
@@ -111,6 +111,13 @@ public abstract class BuildInfoExtractorUtils {
             } else {
                 try (InputStream inputStream = Files.newInputStream(propertiesFile.toPath())) {
                     props.load(inputStream);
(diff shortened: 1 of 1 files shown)


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -111,6 +111,15 @@ private static Properties searchAdditionalPropertiesFile(Properties existingProp
             } else {
                 try (InputStream inputStream = Files.newInputStream(propertiesFile.toPath())) {
                     props.load(inputStream);
+                } catch (IllegalArgumentException e) {
+                    // Handle malformed Unicode escape sequences
```
bhanurp, line 115: All comments looks redundant, consider removing.

PR #824, opened 2025-07-17
RTDEV-58262 - Modify Go pack zip vendor directory

dorsJfrog:
-----


build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoVersionUtils.java
```diff
@@ -15,21 +15,21 @@ public class GoVersionUtils {
...
-    public static long getMajorVersion(String version, Log log) {
+    public static int getMajorVersion(String version, Log log) {
         if (StringUtils.isEmpty(version)) {
             return 0;
         }
         version = getCleanVersion(version);
         Optional<Version> parsedVersion = Version.tryParse(StringUtils.removeStart(version, "v"));
         if (parsedVersion.isPresent()) {
-            return parsedVersion.get().majorVersion();
+            return (int) parsedVersion.get().majorVersion();
```
akushsky, line 32: Why did you switch to int, if `majorVersion()` already returns long?

dorsJfrog: this function used to return int. 
Few months ago @nadav-y used java-semver library the use .majorVersion(); that returns long so it was change.
but artifactory still using the old version of build-info that is expecting for int so its failing on compilation. 
as far as i know, there is no reason for major version to actually be long. so im returning int instead of changing all the refs in artifactory.

nadav-y: why?
why not bump semver version?

for instance, in npm semver could be long. that's why they changed in the the semver lib.
also, blind casting is not a good solution

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -349,4 +364,29 @@ private String getFirstPathElement(String path) {
         path = StringUtils.removeStart(path, "/");
         return StringUtils.substringBefore(path, "/");
     }
+
+    /**
+     * Sets the Go version based on the content of the go.mod file in the root of the zip file.
+     */
+    private void setGoVersionByModFile(ZipArchiveEntry entry) {
+        // Go version 1.12.0 is the default version where we should start looking for go.mod file.
```
BarakHc, line 372: It actually go 1.13 where mod files became mandatory
and go 1.11 where mod file became experimental

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -349,4 +364,29 @@ private String getFirstPathElement(String path) {
...
+                                    modFileContent.split("\n")).filter(line -> line.startsWith("go"))
+                            .findFirst().get().replace("go", "").trim();
+                    Optional<Version> parsedVersion = Version.tryParse(goVersion);
+                    if (!parsedVersion.isPresent()) {
+                        goVersion = goVersion + ".0"; // Ensure the version has a patch.
+                    }
+                    goModVersion = Version.parse(goVersion);
+                }
+            } catch (Exception e) {
+                log.warn("Failed to read go.mod file of the root project: " + ExceptionUtils.getRootCauseMessage(e));
```
BarakHc, line 387: fix the message - it is not just a root project

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -349,4 +364,29 @@ private String getFirstPathElement(String path) {
...
+        if (goModVersion.equals(Version.of(1, 12, 0))) {
+            try (InputStream inputStream = zipFile.getInputStream(entry)) {
+                String modFileContent = IOUtils.toString(inputStream, StandardCharsets.UTF_8.name());
+                if (!StringUtils.isEmpty(modFileContent)) {
+                    String goVersion = Arrays.stream(
+                                    modFileContent.split("\n")).filter(line -> line.startsWith("go"))
+                            .findFirst().get().replace("go", "").trim();
+                    Optional<Version> parsedVersion = Version.tryParse(goVersion);
+                    if (!parsedVersion.isPresent()) {
+                        goVersion = goVersion + ".0"; // Ensure the version has a patch.
```
bhanurp, line 382: it is also possible that from GO 1.21 it can have patch version in go.mod file for example 1.22.1 is also valid make sure .0 is appended only when the go version has major and minor.

https://go.dev/ref/mod#go-mod-file

bhanurp (changes requested):

build-info-extractor-go/src/main/java/org/jfrog/build/extractor/go/extractor/GoZipBallStreamer.java
```diff
@@ -349,4 +364,29 @@ private String getFirstPathElement(String path) {
         path = StringUtils.removeStart(path, "/");
         return StringUtils.substringBefore(path, "/");
     }
+
+    /**
+     * Sets the Go version based on the content of the go.mod file in the root of the zip file.
+     */
+    private void setGoVersionByModFile(ZipArchiveEntry entry) {
+        // Go version 1.13.0 is the version where we should start looking for go.mod file.
+        if (goModVersion.equals(Version.of(1, 12, 0))) {
```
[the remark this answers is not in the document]
dorsJfrog: no we won't read the go.mod file. and we don't need to.
the reason i'v added this to read the go.mod file is due to differences in how we determine a vendor package from go version 1.24.0 . 
so actually for any version that is under that, the behavior will act the same way.
so the default is 1.12.0 only for visibility reasons of go history. it could be 0.0.0 as well but it doesn't matter.

bhanurp (approved):

PR #834, opened 2025-10-17
Reading Extractor Property Temp File Path From System Property

nitinp19:
- We were reading the temp file path from Environmental variable which is causing the issue for the android project as gradle daemon is not respecting the environmental variable. With the new approach we are reading it from the system property which makes it reliable.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
@@ -317,7 +317,7 @@ public abstract class BuildInfoExtractorUtils {
         if ("jar".equals(type)) {
             // add classifier if it exists
             if (StringUtils.isNotBlank(classifier)) {
-                result = classifier + (StringUtils.isNotBlank(extension) && !classifier.endsWith(extension) ? "-" + extension : "");
+                result = classifier;
             }
         }
         // Add extension if not jar, ivy or pom type


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -317,7 +317,7 @@ public static String getTypeString(String type, String classifier, String extens
         if ("jar".equals(type)) {
             // add classifier if it exists
             if (StringUtils.isNotBlank(classifier)) {
-                result = classifier;
+                result = classifier + (StringUtils.isNotBlank(extension) && !classifier.endsWith(extension) ? "-" + extension : "");
```
bhanurp: why is this change required?

build-info-extractor-gradle/src/main/groovy/org/jfrog/gradle/plugin/artifactory/extractor/GradleArtifactoryClientConfigUpdater.java
```diff
@@ -65,6 +89,29 @@ public static void update(ArtifactoryClientConfiguration config, Project project
         addDefaultPublisherAttributes(config, project.getRootProject().getName(), "Gradle", project.getGradle().getGradleVersion());
     }
 
+    /**
+     * Helper method to load build info properties from a file with proper error handling
+     */
+    private static Properties loadBuildInfoProperties(String filePath, org.jfrog.build.api.util.Log log) {
+        if (StringUtils.isBlank(filePath)) {
```
bhanurp: this will always return false since we are creating properties?

nitinp19: yes, this will always be false

PR #838, opened 2025-11-04
excluding ide-specific properties

nitinp19:
Problem- Customer are facing Gradle cache miss happening due to ide properties getting updated in the configuration phase. These properties does not concern the project build but still end up being the one for the cause of cache miss.

Solution- Not reading these properties in the configuration phase which exclude them from being an input in the cache.

### DIFF
--- a/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
+++ b/build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
@@ -102,7 +102,8 @@ public abstract class BuildInfoExtractorUtils {
                 "intellij.",
                 "eclipse.",
                 "netbeans.",
-                "vscode.")) {
+                "vscode.",
+                "studio.")) {
             return false;
         }
         // By default, include all other properties.


build-info-extractor/src/main/java/org/jfrog/build/extractor/BuildInfoExtractorUtils.java
```diff
@@ -77,9 +77,37 @@ public static Properties mergePropertiesWithSystemAndPropertyFile(Properties exi
...
+    
+    /**
+     * This method filters out IDE-specific and other unrelated properties that could cause
+     * configuration cache invalidation in Gradle.
+     */
+    private static boolean isRelevantSystemProperty(String propertyName) {
+        if (StringUtils.isEmpty(propertyName)) {
+            return false;
+        }
+        if (StringUtils.startsWithAny(propertyName,
```
bhanurp: android studio will have the pattern of `android.injected.*` can we add this as well?

nitinp19: sure, let me add it

PR #855, opened 2026-05-04
Fix Maven 3.9.12 + Develocity compatibility: declare checkPrerequisites override

galusben:
## Summary
- Maven 3.9.12 renamed `DefaultMavenPluginManager#checkRequiredMavenVersion` to `checkPrerequisites`. The Develocity Maven extension reflectively looks the new method up via `Class.getDeclaredMethod` on the Plexus-bound `DefaultMavenPluginManager` implementation, which in this project is `ArtifactoryEclipsePluginManager`. `getDeclaredMethod` does not walk the superclass, so the lookup throws `NoSuchMethodException` and blocks any Maven >= 3.9.12 build that also loads Develocity's extension.
- PR #844 (in v2.43.6) added the `prerequisitesCheckers` Plexus field requirement, which addressed the NPE variant on Maven 3.9.12 — but did not declare the method, so the Develocity path stayed broken.
- This PR bumps `maven-core` build dep from 3.8.6 to 3.9.12 so the override compiles cleanly, declares `checkPrerequisites` on the subclass (delegating to `super`), and adds a regression unit test that mirrors Develocity's exact reflective lookup.

## Failing stack trace this fixes (from a real CI log)
```
java.lang.NoSuchMethodException: org.jfrog.build.extractor.maven.resolver.ArtifactoryEclipsePluginManager.checkPrerequisites(org.apache.maven.plugin.descriptor.PluginDescriptor)
    at java.lang.Class.getDeclaredMethod (Class.java:2422)
    at com.gradle.maven.cache.extension.g.b.checkPrerequisites (SourceFile:117)
```

## Repro confirmed locally
Loading the master HEAD uber jar and running `Class.forName("…ArtifactoryEclipsePluginManager").getDeclaredMethod("checkPrerequisites", PluginDescriptor.class)` reproduces the exact same `NoSuchMethodException`. With this branch, the lookup returns the declared method and the exception goes away.

## Test plan
- [x] New `ArtifactoryEclipsePluginManagerTest.testCheckPrerequisitesIsDeclaredOnSubclass` mirrors Develocity's exact reflective contract — fails on master, passes here.
- [x] `./gradlew :build-info-extractor-maven3:compileJava :build-info-extractor-maven3:test` clean against JDK 11 + the bumped maven-core 3.9.12 build dep.
- [ ] Suggested follow-up (not in this PR): extend `.github/workflows/maven-compatibility-test.yml` to install Develocity's Maven extension in the test pom and fail on `NoSuchMethodException`, so this regression class is caught by the nightly job.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
@@ -31,9 +31,8 @@ public class ArtifactoryEclipsePluginManager extends DefaultMavenPluginManager {
         return super.setupExtensionsRealm(project, plugin, session);
     }
 
-    // Declared so Develocity's reflective lookup
-    // (Class.getDeclaredMethod("checkPrerequisites", PluginDescriptor.class)) succeeds —
-    // getDeclaredMethod does not walk the superclass.
+    // Declared so Class.getDeclaredMethod on this subclass finds it —
+    // getDeclaredMethod does not walk the superclass chain.
     @Override
     public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginIncompatibleException {
         super.checkPrerequisites(pluginDescriptor);
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
@@ -13,6 +13,9 @@ import org.codehaus.plexus.component.annotations.Requirement;
 import org.eclipse.aether.RepositorySystemSession;
 
 import javax.inject.Named;
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
 import java.util.List;
 
 /**
@@ -21,6 +24,23 @@ import java.util.List;
 @Named
 @Component(role = DefaultMavenPluginManager.class)
 public class ArtifactoryEclipsePluginManager extends DefaultMavenPluginManager {
+
+    // Non-virtual handle to DefaultMavenPluginManager#checkPrerequisites if and only if the
+    // parent class declares it (Maven >= 3.9.12). On older Maven runtimes the method does not
+    // exist on the parent, the handle stays null, and the override below is a safe no-op.
+    private static final MethodHandle SUPER_CHECK_PREREQUISITES = resolveSuperCheckPrerequisites();
+
+    private static MethodHandle resolveSuperCheckPrerequisites() {
+        try {
+            return MethodHandles.lookup().findSpecial(
+                    DefaultMavenPluginManager.class,
+                    "checkPrerequisites",
+                    MethodType.methodType(void.class, PluginDescriptor.class),
+                    ArtifactoryEclipsePluginManager.class);
+        } catch (NoSuchMethodException | IllegalAccessException e) {
+            return null;
+        }
+    }
 
     @Requirement
     private ArtifactoryEclipseResolversHelper helper;
@@ -32,10 +52,22 @@ public class ArtifactoryEclipsePluginManager extends DefaultMavenPluginManager {
     }
 
     // Declared so Class.getDeclaredMethod on this subclass finds it —
-    // getDeclaredMethod does not walk the superclass chain.
+    // getDeclaredMethod does not walk the superclass chain. Delegates to the parent via a
+    // MethodHandle resolved at class-load time so the bytecode does not embed a hard reference
+    // to a method that may not exist on older Maven runtimes (3.8.x / 3.9.0 - 3.9.11).
     @Override
     public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginIncompatibleException {
-        super.checkPrerequisites(pluginDescriptor);
+        MethodHandle handle = SUPER_CHECK_PREREQUISITES;
+        if (handle == null) {
+            return;
+        }
+        try {
+            handle.invoke(this, pluginDescriptor);
+        } catch (PluginIncompatibleException | RuntimeException | Error e) {
+            throw e;
+        } catch (Throwable t) {
+            throw new RuntimeException(t);
+        }
     }
 
     private void enforceResolutionRepositories(MavenProject project, RepositorySystemSession session) {
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
@@ -3,7 +3,9 @@ package org.jfrog.build.extractor.maven.resolver;
 import org.apache.maven.artifact.repository.ArtifactRepository;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.ExtensionRealmCache;
+import org.apache.maven.plugin.PluginIncompatibleException;
 import org.apache.maven.plugin.PluginManagerException;
+import org.apache.maven.plugin.descriptor.PluginDescriptor;
 import org.apache.maven.plugin.internal.DefaultMavenPluginManager;
 import org.apache.maven.project.MavenProject;
 import org.codehaus.plexus.component.annotations.Component;
@@ -27,6 +29,14 @@ public class ArtifactoryEclipsePluginManager extends DefaultMavenPluginManager {
     public ExtensionRealmCache.CacheRecord setupExtensionsRealm(MavenProject project, Plugin plugin, RepositorySystemSession session) throws PluginManagerException {
         enforceResolutionRepositories(project, session);
         return super.setupExtensionsRealm(project, plugin, session);
+    }
+
+    // Declared so Develocity's reflective lookup
+    // (Class.getDeclaredMethod("checkPrerequisites", PluginDescriptor.class)) succeeds —
+    // getDeclaredMethod does not walk the superclass.
+    @Override
+    public void checkPrerequisites(PluginDescriptor pluginDescriptor) throws PluginIncompatibleException {
+        super.checkPrerequisites(pluginDescriptor);
     }
 
     private void enforceResolutionRepositories(MavenProject project, RepositorySystemSession session) {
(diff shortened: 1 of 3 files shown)


galusben:
## Root cause and where this should be fixed

Three parties are involved; each one is doing something a little brittle.

**1. Maven made a backwards-incompatible API change in a patch release.**
Verified with `javap` on `maven-core`:
- 3.9.11 declares `public void checkRequiredMavenVersion(PluginDescriptor)`
- 3.9.12 renames it to `public void checkPrerequisites(PluginDescriptor)` (the new name reflects the related extension point added in 3.9.12 — `prerequisitesCheckers`).

**2. The Develocity Maven extension uses `Class.getDeclaredMethod(...)` for the lookup.** `getDeclaredMethod` only returns methods physically declared on the class you call it on; it does *not* walk the superclass chain. `Class.getMethod(...)` would walk inheritance for public methods.

**3. `ArtifactoryEclipsePluginManager` is registered (via Plexus `components.xml`) as the implementation of `DefaultMavenPluginManager`.** Develocity therefore reflects against *this subclass*, not Maven's class. The subclass only directly declares `setupExtensionsRealm`, so every other method is inherited — and `getDeclaredMethod("checkPrerequisites", …)` on the subclass throws `NoSuchMethodException`.

In Maven 3.9.11 the same anti-pattern would have applied to `checkRequiredMavenVersion`, but Develocity wasn't reflecting on that method back then. They presumably added the reflective hook *because* of the 3.9.12 rename, and the moment they did, the latent issue with our subclass surfaced.

### Why fix it here

Develocity could harden the reflective lookup (use `getMethod`, or fall back when `getDeclaredMethod` misses). That's the more durable fix and we should share this PR with them. But:
- We don't ship Develocity, so we can't unblock our users by waiting on a Develocity release.
- The change is small and self-contained: a `MethodHandle` lookup at class-load time plus a thin override that delegates to whichever method name the running Maven actually has, plus a regression test.
- It mirrors the precedent set by #844 (in v2.43.6), which kept up with the *field* side of the same Maven 3.9.12 reorg (`prerequisitesCheckers`). This PR is the missing companion piece for the *method* side.

If/when Develocity loosens their lookup, this override becomes a no-op safety net rather than load-bearing.

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
```diff
@@ -29,6 +31,13 @@ public ExtensionRealmCache.CacheRecord setupExtensionsRealm(MavenProject project
         return super.setupExtensionsRealm(project, plugin, session);
     }
 
+    // Declared so Class.getDeclaredMethod on this subclass finds it —
+    // getDeclaredMethod does not walk the superclass chain.
+    @Override
```
agrasth, line 72: `checkPrerequisites` doesn't exist in `DefaultMavenPluginManager` before Maven `3.9.12`. On Maven `3.8.x` without Develocity this is fine (method is never called). But with Develocity, `getDeclaredMethod` now finds this override and invokes it — causing `super.checkPrerequisites()` to throw `NoSuchMethodError` at runtime. 
This replaces one error with another for users still on Maven 3.8.x + Develocity.

build-info-extractor-maven3/src/test/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManagerTest.java
```diff
@@ -0,0 +1,23 @@
+package org.jfrog.build.extractor.maven.resolver;
+
+import org.apache.maven.plugin.descriptor.PluginDescriptor;
+import org.testng.annotations.Test;
+
+import java.lang.reflect.Method;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
```
agrasth: This test runs under TestNG (`@org.testng.annotations.Test`) but uses JUnit 4 assertions. This only works incidentally because TestNG has an optional transitive dependency on `junit:4.13.2`. Please use TestNG's own assertions instead:
```
import static org.testng.Assert.assertEquals;
import static org.testng.Assert.assertNotNull;
```

build.gradle
```diff
@@ -26,7 +26,7 @@ buildscript {
         jacksonVersion = '2.17.3'
         jdomVersion = '2.0.6.1'
         jsrVersion = '3.0.2'
-        mavenVersion = '3.8.6'
+        mavenVersion = '3.9.12'
```
agrasth, line 29: mavenVersion is used in 3 places, not just maven-core:
```
  // build.gradle lines 492-503
  "org.apache.maven:maven-repository-metadata:$mavenVersion",  // bumped 3.8.6 → 3.9.12
  "org.apache.maven:maven-model:$mavenVersion",                // bumped 3.8.6 → 3.9.12
  ...
  implementation("org.apache.maven:maven-core:$mavenVersion")  // bumped 3.8.6 → 3.9.12
```
PR intended to only bump maven-core to get access to checkPrerequisites, but it accidentally also bumped maven-repository-metadata and maven-model as a side effect since they share the same variable.
maybe we can split the variable:
```
  mavenVersion = '3.8.6'         // for maven-repository-metadata, maven-model
  mavenCoreVersion = '3.9.12'    // for maven-core only
```
wdyt @bhanurp ?

galusben:
## CI triage — none of the red checks are caused by this PR

Compared this PR's runs against the master HEAD CI run (`24656256484`, sha `013da9b0`).

**Pre-existing on master HEAD (and in 5 of 5 recent master pushes):**
- ❌ Distribution (3 OSes) — `Bad credentials` on `setUp` of `*DistributionManagerTest`
- ❌ Go (3 OSes) — `401: Wrong username was used` on `GoExtractorTest.setUp`
- ❌ Gradle (ubuntu, windows) — `UnexpectedBuildFailure` running `artifactoryPublish` in testkit
- ❌ Extractor (macos-latest) — also failed on master HEAD

**Initially red on this PR but green on rerun (transient flakes, coincided with a GitHub Actions cache outage — `"Our services aren't available right now"`):**
- ✅ Docker — was `unknown blob` / `gcr.io` pull failures; now passing
- ✅ Python (windows-latest) — was a contaminated pip venv on the windows runner; now passing

**Stable green across both runs — the jobs this change actually exercises:**
- ✅ Maven (ubuntu / macos / windows)
- ✅ Extractor (ubuntu / windows) — runs the new `ArtifactoryEclipsePluginManagerTest` regression test
- ✅ Ivy (3 OSes), npm, NuGet, Python (ubuntu)

The mavenVersion bump (3.8.6 → 3.9.12) is consumed only by `build-info-extractor-maven3`. Distribution / Go / Gradle / Docker / pip don't pull `maven-core` and the test surface that touched our change (Maven matrix + maven3 unit tests) is fully green.

build.gradle
galusben: I would keep the single version for simplicity.
On some point you want to upgrade everything.

build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManager.java
galusben: Good catch — fixed in 6a24180d. Switched to a `MethodHandle.findSpecial` lookup resolved at class-load time so the bytecode no longer embeds a hard `invokespecial` reference to a method that may not exist on the parent.

- Maven ≥ 3.9.12: handle is non-null → our override delegates to super.
- Maven < 3.9.12 (3.8.x / 3.9.0 – 3.9.11): static initializer catches `NoSuchMethodException` from `findSpecial`, handle stays null, the override returns a no-op. No `NoSuchMethodError` at invoke time.

Verified locally by loading the same uber jar against both Maven 3.9.11 maven-core and 3.9.12 maven-core on the classpath:

```
=== Maven 3.9.11 (no checkPrerequisites on parent) ===
STEP 1 OK: getDeclaredMethod returned ...checkPrerequisites...
STEP 2 OK: checkPrerequisites returned without NoSuchMethodError

=== Maven 3.9.12 (parent has checkPrerequisites) ===
STEP 1 OK: getDeclaredMethod returned ...checkPrerequisites...
STEP 2 OK: super was invoked, threw NullPointerException (expected on 3.9.12+ with empty descriptor)
```

build-info-extractor-maven3/src/test/java/org/jfrog/build/extractor/maven/resolver/ArtifactoryEclipsePluginManagerTest.java
galusben: Done in 6a24180d — swapped to TestNG's `assertEquals` / `assertNotNull` and flipped the argument order to TestNG's `(actual, expected)` convention. Note: the existing `ArtifactoryResolutionTest` in the same package uses JUnit assertions too, so we may want a follow-up to clean that up — but I left it out of this PR's scope.

galusben:
## Update: also delegate on Maven < 3.9.12 (per @galusben suggestion)

@galusben raised: when called by Develocity on Maven < 3.9.12, the previous version of the override silently no-op'd — bypassing whatever prerequisites check the caller expected. New commit `97153646` makes the static initializer also resolve `checkRequiredMavenVersion` (the pre-3.9.12 name), so the check runs on every supported Maven version.

### Subtle trap worth flagging
First attempt looked up the old name *first*, falling back to the new one. That hit a deprecation shim: on Maven 3.9.12 the `MavenPluginManager` interface keeps `checkRequiredMavenVersion` as a **deprecated default method** that delegates back to `checkPrerequisites` via virtual dispatch — which lands on *our override* — infinite recursion → `StackOverflowError`.

Verified with `javap`:
```
# 3.9.11 interface
public abstract void checkRequiredMavenVersion(PluginDescriptor)

# 3.9.12 interface
public default void checkRequiredMavenVersion(PluginDescriptor)   ← deprecation shim
public default void checkPrerequisites(PluginDescriptor)
```

Fix: try the **new name first**. On 3.9.12 we hit the real concrete impl on `DefaultMavenPluginManager` directly and never touch the shim. On < 3.9.12 the new name doesn't exist; we fall back to the old name, which is the real concrete impl on those versions.

### End-to-end verification

Same uber jar, two Maven runtimes:

```
=== Maven 3.9.11 ===
STEP 1 OK: getDeclaredMethod returned ...checkPrerequisites...
STEP 2 super was invoked, threw NPE
   at DefaultMavenPluginManager.checkRequiredMavenVersion(DefaultMavenPluginManager.java:293)  ← real parent method
   at ArtifactoryEclipsePluginManager.checkPrerequisites(ArtifactoryEclipsePluginManager.java:78)

=== Maven 3.9.12 ===
STEP 1 OK: getDeclaredMethod returned ...checkPrerequisites...
STEP 2 super was invoked, threw NPE
   at DefaultMavenPluginManager.checkPrerequisites(DefaultMavenPluginManager.java:290)         ← real parent method
   at ArtifactoryEclipsePluginManager.checkPrerequisites(ArtifactoryEclipsePluginManager.java:78)
```

The NPE is from the parent touching uninitialized state on a stand-alone `PluginDescriptor` (outside Maven's container) — what matters is *which* parent method ran. On both versions, the call lands in the right place.

agrasth (approved):

PR #862, opened 2026-09-06
RTECO-1424: Default to Maven Unique Snapshot Timestamp in the Extractor Flow

naveenku-jfrog:
### Problem
`jf mvn deploy` runs through the Build Info Extractor, which sets `maven.deploy.skip=true` and uploads artifacts itself at session end. Because `maven-deploy-plugin` never runs, its snapshot-version transformation never fires — so SNAPSHOT artifacts were deployed under the non-unique base name (`my-app-1.0-SNAPSHOT.jar`), regardless of `-DuniqueVersion` or the repository's *Maven Snapshot Version Behavior*. This diverges from native Maven 3, which only produces unique timestamped snapshots. Customers forced onto the extractor flow (e.g. a `maven.yaml` present, which pins the extractor even with `JFROG_RUN_NATIVE=true`) could not get timestamped snapshots without workarounds.

### Fix
Reproduce Maven 3's snapshot naming from inside the extractor:

- **`SnapshotVersionResolver`** builds `<base>-<yyyyMMdd.HHmmss>-<buildNumber>` from one session-wide UTC timestamp, with the build number read from the repository's snapshot `maven-metadata.xml` (falls back to `1` when unreadable, keeping the timestamp). The result is cached per `group:artifactId:version`, so a module's jar / pom / attached artifacts all share one version.
- **`BuildInfoRecorder`** resolves the deployed version when collecting artifacts and threads it through the artifact name and deployment path. The directory stays at the `-SNAPSHOT` path, so snapshot-repo routing is unchanged; only the file name carries the timestamp.
- **Defaults to unique** (matching Maven 3). Honors the deploy plugin's own `-DuniqueVersion=false` to opt out, or the publisher's new `snapshot.unique` configuration when there is no command line to set it on.
- Adds the `snapshot.unique` client-configuration field (default `true`).

### Tests
`SnapshotVersionResolverTest` covers release passthrough, the `uniqueVersion=false` opt-out, first deployment, build-number-from-metadata, and one shared version across a module. `:build-info-extractor-maven3:test` is green.

### Verification
Verified end-to-end through the real `jf mvn` CLI against Artifactory, using a snapshot repo in **Deployer** mode:

| Scenario | Deployed name |
|---|---|
| Released `2.43.9` jar (before) | `my-app-1.0-SNAPSHOT.jar` |
| Fixed jar, default | `my-app-1.0-20260906.073650-1.jar` |
| Fixed jar, `-DuniqueVersion=false` | `my-app-1.0-SNAPSHOT.jar` |

### Downstream follow-up
After this is released, bump `MavenExtractorDependencyVersion` in `build-info-go` (`build/maven.go`) so `jf mvn` downloads the fixed extractor.

### DIFF
--- a/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
+++ b/build-info-extractor-maven3/src/main/java/org/jfrog/build/extractor/maven/BuildInfoRecorder.java
@@ -62,6 +62,11 @@ import static org.jfrog.build.extractor.clientConfiguration.ClientConfigurationF
 @Component(role = BuildInfoRecorder.class)
 public class BuildInfoRecorder extends AbstractExecutionListener implements BuildInfoExtractor<ExecutionEvent> {
 
+    /**
+     * maven-deploy-plugin's own flag for choosing between unique and non-unique snapshots.
+     */
+    private static final String UNIQUE_VERSION_USER_PROPERTY = "uniqueVersion";
+
     @Requirement
     private BuildInfoModelPropertyResolver buildInfoModelPropertyResolver;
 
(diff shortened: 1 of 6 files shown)


agrasth:
We can close this PR, as discussed. We will not move ahead with this approah/fix.

